PR #14, opened 2020-10-26
175192569 smsfixes

davkutalek:
Updates to use more universal SMS intent api and an SMS observer to detect when SMS are actually sent.


androidmaven:
LGTM

PR #17, opened 2020-11-04
Request new account

androidmaven:


app/src/main/java/com/hover/stax/requestAccount/RequestAccountViewModel.java
```diff
@@ -0,0 +1,103 @@
...
+	public void sendContactInfoToFirebase(String phone, String email, String deviceId) {
+		Map<String, Object> map = new HashMap<>();
+		map.put("deviceId", deviceId);
+		map.put("phone", phone);
+		map.put("email", email);
+		map.put("timeStamp", Timestamp.now());
+		FirebaseFirestore.getInstance().collection("users_contact").add(map);
+	}
+
+	private List<SupportedCountries> getSupportedCountries() {
```
davkutalek, line 93: This should be in a repo class which downloads options from Hover's API and communicates the finished request to firebase

davkutalek (approved):
This is great. The AccountsByCountry and Supported Countries objects are a little strange. The data should be pulled from the same Hover APIs that our web dashboard uses. Everything else is ready to go, I'm going to merge this but keep the pivotal story open.

PR #20, opened 2020-11-30
Update strings.xml

justinscherer:
Made a few changes based on user testing. Not sure if you need to double check these at all @davkutalek


davkutalek (approved):
Thanks!

PR #22, opened 2020-12-06
Payment link clean up

androidmaven:
Read some key points in the clean-code's book. Realized a lot of violations I've been making, and practised correcting them.

Keypoints I learnt for the refractor of the Channel's folder.

Arguments should be input; output arguments should be avoided.
Use verbs for method name; nouns for classes.
Functions should do only one thing or sectioned steps.
Methods, arguments and variable naming should be descriptive and contextualized as possible.

### DIFF
--- a/app/src/main/java/com/hover/stax/channels/Channel.java
+++ b/app/src/main/java/com/hover/stax/channels/Channel.java
@@ -12,7 +12,9 @@ import com.hover.stax.utils.DateUtils;
 import org.json.JSONException;
 import org.json.JSONObject;
 
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 
 @Entity(tableName = "channels")
 public class Channel {
@@ -115,6 +117,15 @@ public class Channel {
 		} else {
 			latestBalanceTimestamp = DateUtils.now();
 		}
+	}
+
+	static List<Channel> getCountryChannels(String countryAlpha2, List<Channel> channels) {
+		List<Channel> countryChannels = new ArrayList<>();
+		for (int i = 0; i < channels.size(); i++) {
+			if (countryAlpha2.equals(channels.get(i).countryAlpha2.toUpperCase()))
+				countryChannels.add(channels.get(i));
+		}
+		return countryChannels;
 	}
 
 	@Override
--- a/app/src/main/java/com/hover/stax/channels/ChannelListFragment.java
+++ b/app/src/main/java/com/hover/stax/channels/ChannelListFragment.java
@@ -46,6 +46,11 @@ public class ChannelListFragment extends Fragment implements ChannelsAdapter.Sel
 	}
 
 	private void addChannels(View view) {
+		observeToSetupChannelViews(view);
+		observeToSetupCountryChannels(view);
+		observeToSetupAllChannels(view);
+	}
+	private void observeToSetupChannelViews(View view) {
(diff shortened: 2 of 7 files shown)


app/src/main/java/com/hover/stax/channels/ChannelListViewModel.java
```diff
@@ -42,9 +42,10 @@ public ChannelListViewModel(Application application) {
 		super(application);
 		repo = new DatabaseRepo(application);
 		loadChannels();
+		initSelectedChannels();
 		loadSims();
-		simHniList = Transformations.map(sims, this::getHnis);
-		simCountryList = Transformations.map(sims, this::getSimCountries);
+		simHniList = Transformations.map(sims, this::getHnisAndSubscribeToEachOnFirebase);
```
davkutalek, line 47: Nice. I do tend to go for shorter names over fully descriptive names cause I am lazy, but this is a good example of why that is bad and this is much better.

app/src/main/java/com/hover/stax/channels/ChannelsActivity.java
```diff
@@ -63,12 +63,12 @@ private void goToChannelSelection() {
 		findViewById(R.id.continue_btn).setVisibility(View.VISIBLE);
 	}
 
-	private void onSelectedUpdate(List<Integer> ids) {
-		if (ids.size() > 0) {
+	private void validateToLogeventThenSaveAndContinue(List<Integer> channelIds) {
```
davkutalek, line 66: This is an improvement, but the really right thing would be to name this function something like `changeContinueClickAction`, then move the actual click action to their own functions as well, something like `findViewById(R.id.continue_btn).setOnClickListener(view -> logChoicesSaveAndContinue)`

davkutalek (approved):
This is great, practicing is a really good idea. These changes are good, and in general I find your naming to be quite good already. I personally really need to practice avoiding output arguments, though I blame Android a little.

app/src/main/java/com/hover/stax/channels/Channel.java
```diff
@@ -117,6 +119,15 @@ public void updateBalance(HashMap<String, String> parsed_variables) {
 		}
 	}
 
+	static List<Channel> getCountryChannels(String countryAlpha2, List<Channel> channels) {
```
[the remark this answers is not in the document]
androidmaven: You're right, the best thing I think is to move the function back to the fragment.

app/src/main/java/com/hover/stax/channels/ChannelListFragment.java
```diff
@@ -46,6 +46,11 @@ public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceStat
 	}
 
 	private void addChannels(View view) {
+		observeToSetupChannelViews(view);
```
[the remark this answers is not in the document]
androidmaven: Uhmm, Changing to `setUpChannelObservers` is rightly better. 

But `setUpChannelViews` uses arguments View view and Channel channel; which means we'd have to pass in two arguments; of which 1 argument is better than 2. 

To be able to pass in one argument, we'd have to initialize View outside the function, which i think isn't enough justification.

app/src/main/java/com/hover/stax/channels/ChannelsActivity.java
androidmaven: Yup, you're right

davkutalek:
Merge directly to development

PR #24, opened 2020-12-14
completed making adding pin invisible and new showcase update

androidmaven:
- Removed PIN management from settings.
- Go from adding accounts back to home page, instead of showing the PIN screen
- Make add account showcase show whenever the balance is empty;
- Make remaining showcase run after adding accounts.

### DIFF
--- a/app/src/main/java/com/hover/stax/channels/ChannelsActivity.java
+++ b/app/src/main/java/com/hover/stax/channels/ChannelsActivity.java
@@ -85,8 +85,14 @@ public class ChannelsActivity extends AppCompatActivity {
 	private void saveAndContinue() {
 		channelViewModel.saveSelected();
 		Utils.saveInt(AUTH_CHECK, 1, this);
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/hover/stax/channels/ChannelsActivity.java
```diff
@@ -85,8 +85,14 @@ private void logChoicesSaveAction(List<Integer> channelIds) {
 	private void saveAndContinue() {
 		channelViewModel.saveSelected();
 		Utils.saveInt(AUTH_CHECK, 1, this);
-		startActivityForResult(new Intent(ChannelsActivity.this, PinsActivity.class), 0);
+		goToMainActivity();
+		//startActivityForResult(new Intent(ChannelsActivity.this, PinsActivity.class), 0);
 	}
+	private void goToMainActivity() {
```
davkutalek, line 91: Not necessarily returning to Main, so shouldn't be named like this

davkutalek (approved):
LGTM

PR #29, opened 2020-12-18
176209172 minor UI bug fixes

androidmaven:


davkutalek (commented):
Creating v2 of classes is bad! It indicates that the work to properly integrate everything has not been done and leaves no explanation for why not. I know changing every file is a pain, but there are tools that can help.

PR #30, opened 2021-01-06
Jan6 crashlytics fixes attempts

androidmaven:


app/src/main/java/com/hover/stax/utils/paymentLinkCryptography/ZRMClass.java
```diff
@@ -1,6 +1,7 @@
 package com.hover.stax.utils.paymentLinkCryptography;
 
 class ZRMClass {
+	//TODO
```
davkutalek, line 4: Why is this class here? TODOs should usually be avoided, but should at least include some description of what/why. If it really needs doing it should ideally go in Pivotal.

davkutalek (changes requested):
Minor changes requested then go ahead and merge it

PR #32, opened 2021-02-04
added migration snippets with jornal mode non-destructive

androidmaven:
1. Checked out from the development branch.


davkutalek:
@androidmaven mind writing a quick explanation of why you switched the Journal mode to truncate? Just curious cause I previously had to set it to write ahead to make weird loading failures stop. I don't really understand it, so if you have any sources I could read that would help.

PR #35, opened 2021-02-12
Sprint2

androidmaven:
- Bottom nav bar showing on all screens.
- New link account fragment.


app/src/main/java/com/hover/stax/utils/errors/NoticeDrawable.java
```diff
@@ -0,0 +1,20 @@
+package com.hover.stax.utils.errors;
+
+import com.hover.stax.R;
+
+public class NoticeDrawable {
```
davkutalek, line 5: This should be removed from this PR I believe

davkutalek (changes requested):
Overall code looks good. Requesting minor changes to the way the interface/link event works in Channel Dropdown. 

In terms of functionality couple small things as well: 
1. Nav always shows above keyboard. Maybe double check with Justin, but I think it should be hidden by keyboard. 
2. When visiting the new link account fragment the dropdown starts with a value selected but it should be empty. 
3. Once selected, when pressing link account only the new account balance should be checked, not all.

app/src/main/java/com/hover/stax/settings/PinsViewModel.java
```diff
@@ -81,14 +81,14 @@ public void removeAccount(Channel channel) {
 		channel.selected = false;
 		channel.defaultAccount = false;
 		repo.update(channel);
-		if (channels.getValue() == null || !changeDefault) return;
+		/*if (channels.getValue() == null || !changeDefault) return;
```
davkutalek: Why is this commented?

davkutalek (approved):
LGTM, just one question

app/src/main/java/com/hover/stax/settings/PinsViewModel.java
androidmaven: I've uncommented it back. It wasn't removing the last account. I understand it's trying to set the last standing account as the default account; so tagging it as a bug and added a TODO to the latest commit to fix this my morning tomorrow.

PR #36, opened 2021-02-23
Sprint3

androidmaven:
Issue I'm facing is that it seems only one view can be on the screen's focus at a time. And for the border to remain green, it needs to stay in focus.


androidmaven:
Issues resolved.

davkutalek (approved):
This is ok but could use a little refactoring, see comments

davkutalek:
Looking at this in more detail, it has serious problems. Lets chat about it tomorrow

app/src/main/java/com/hover/stax/balances/BalancesFragment.java
```diff
@@ -105,12 +105,17 @@ private void updateServices(List<Channel> channels, View view) {
 	}
 
 	private void refreshBalances(View v) {
-		if (channelDropdown.getHighlighted() != null) {
+		if (channelDropdown.getHighlighted() != null
```
davkutalek, line 108: This is not correct. The `setAllRunning` function already causes the view model to fire once the actions have updated if there are none, so setting an observer here is redundant and seriously confuses the already complicated logic in this function.

This should work fine as is. If you don't think it is working please write a failing test first so that we can verify there is a problem.

app/src/main/java/com/hover/stax/transfers/TransferFragment.java
```diff
@@ -206,11 +212,22 @@ protected void onContactSelected(int requestCode, StaxContact contact) {
 		}
 	};
 
-	private void load(Request r) {
+	private void loadAndIndicateFieldState(Request r) {
```
davkutalek, line 215: This name is much worse. The field state is secondary effect of loading the request. As the Clean Code book puts it, it is a different level of abstraction. If I'm reading this code with no idea what is going on I can understand what load(Request) means but I have no idea what a field state is.

This is a code smell for me that tells me something weird is going on. Why would we set the fieldstate manually for a request, compared to anything else?

app/src/main/java/com/hover/stax/transfers/TransferFragment.java
```diff
@@ -206,11 +212,22 @@ protected void onContactSelected(int requestCode, StaxContact contact) {
...
 		amountInput.setText(r.amount);
 		recipientAutocomplete.setText(r.requester_number);
 		transferViewModel.setEditing(r.amount == null || r.amount.isEmpty());
+		indicateFieldState(r.amount, r.requester_number);
 		Amplitude.getInstance().logEvent(getString(R.string.loaded_request_link));
 	}
+	private void indicateFieldState(String amount, String requesterNum) {
+		if(amount == null || amount.isEmpty()) amountEntry.requestFocus();
+		if(requesterNum !=null && !requesterNum.isEmpty()) {
+			new Handler().postDelayed(() -> {
```
davkutalek, line 226: Definitely don't this. Why is there a delay? Why does it need a thread? Why does this only happen if the requester number exists? Why are we setting the state manually here, but we aren't setting the success state for channel dropdown in any other case? Why does the setSuccess method take an empty String at all, particularly since it is never used any other way?

app/src/main/java/com/hover/stax/views/CustomTextInputLayout.java
```diff
@@ -0,0 +1,120 @@
...
+	public CustomTextInputLayout(@NonNull Context context, @Nullable AttributeSet attrs) {
+		super(context, attrs);
+
+		getAttrs(context,attrs);
+		LayoutInflater.from(context).inflate(R.layout.custom_textinputlayout, this);
+		initViews();
+		fillAttr();
+	}
+
+	//INITIALIZATIONS
```
davkutalek, line 34: Why are there useless comments?

davkutalek (changes requested):
See new comments for things that need changes. I know you can do better! Please respond to each comment and let me know if you want to chat.

app/src/main/java/com/hover/stax/transfers/TransferFragment.java
androidmaven: You're right about it being a different level of abstraction. But I put this here because of the direct effect of loading a request. Based on the design, how I understood and implemented the logic was:
- if a user wants to fulfil a request; Account dropdown, and account number fields should turn green by default without users input, so far as the data are not empty. 

- This is currently the only scenario where a field visibly shows success field state, which was why I set the fieldstate manually

app/src/main/java/com/hover/stax/transfers/TransferFragment.java
androidmaven: Handler shouldn't be here, I used this while trying to figure out why channelDrop state kept changing while loading. And got the comment on field state not existing as its own object. Fix coming in next commit

app/src/main/java/com/hover/stax/views/CustomTextInputLayout.java
androidmaven: Added these comments, because while the methods in the class are descriptive, finding certain methods amongst a bunch of others, was easier when the methods were structured in grouped purpose - > initialization, methods dealing with states, and private methods used within the class. But maybe removing comments and just add decent spacing will do.

app/src/main/java/com/hover/stax/balances/BalancesFragment.java
androidmaven: Very costly oversight. Apologies. It actually works without the observer, forgot to remove it.

app/src/main/java/com/hover/stax/views/CustomTextInputLayout.java
```diff
@@ -0,0 +1,120 @@
...
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import com.google.android.material.textfield.TextInputEditText;
+import com.google.android.material.textfield.TextInputLayout;
+import com.hover.stax.R;
+import com.hover.stax.utils.Utils;
+
+public class CustomTextInputLayout extends TextInputLayout {
```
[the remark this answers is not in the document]
androidmaven: I created a separate object for dropdown and text; cos of:
1. Different styling.
2. Different View type inside the TextInputLayout. Dropdown uses Autocomplete, while text uses an edit. 

I thought of it being cleaner when separated for anyone reading the code; and dropdown functionality could be extended for whatever reason without having it affect texts.

app/src/main/java/com/hover/stax/channels/ChannelDropdown.java
```diff
@@ -26,7 +28,7 @@
 public class ChannelDropdown extends TextInputLayout implements Target {
```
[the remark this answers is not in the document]
androidmaven: It should. Will fix in next commit

app/src/main/java/com/hover/stax/fieldstates/FieldState.java
```diff
@@ -0,0 +1,27 @@
+package com.hover.stax.fieldstates;
+
+public class FieldState {
```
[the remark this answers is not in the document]
androidmaven: It should. Will fix in next commit

app/src/main/java/com/hover/stax/fieldstates/FieldStateType.java
```diff
@@ -0,0 +1,6 @@
+package com.hover.stax.fieldstates;
+
+public enum FieldStateType {
```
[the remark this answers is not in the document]
androidmaven: I thought of it that, enum was a better use here cos:

- It houses the types of field state we have and helps in code readability. Instead of passing integers that could accept invalid data.

app/src/main/java/com/hover/stax/views/CustomDropdownLayout.java
```diff
@@ -0,0 +1,147 @@
...
+import android.widget.ImageView;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import com.google.android.material.textfield.TextInputLayout;
+import com.hover.stax.R;
+import com.hover.stax.utils.Utils;
+
+public class CustomDropdownLayout extends TextInputLayout {
```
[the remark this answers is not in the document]
androidmaven: This layout is used only for both - dropdowns and selection of contact (which also has ability to option of selecting from suggestions). Still unsure of another way to name it, CustomInputLayout /StaxInputLayout  may be best describe what it does too

app/src/main/java/com/hover/stax/fieldstates/FieldState.java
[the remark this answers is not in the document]
androidmaven: I went through this again. I thought of this object to serve both the CustomDropdown and CustomTextInputLayout. Reason why both dropdown and textinput isn't using the same custom layout is because I haven't found a way to set styling programmatically for it.

app/src/main/java/com/hover/stax/views/CustomDropdownLayout.java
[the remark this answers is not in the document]
davkutalek: Try to keep the naming consistent. We use Stax... everywhere else so do that here too.

app/src/main/java/com/hover/stax/views/CustomTextInputLayout.java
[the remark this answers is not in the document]
davkutalek: It should work like the base material design TextInputLayout so that you can pass either a EditText or AutoComplete as the child depending on which you need. That is a serious reduction in code duplication and I don't believe there are any style differences between the TextInput part, the child is where the style differences are.

I did actually make an attempt to combine these to see if it is possible, but ran into issues with the children being added in the right place, so if it is too much work to combine then having different classes is fine for now.

app/src/main/java/com/hover/stax/transfers/TransferFragment.java
davkutalek: Ah, gotcha. That is not correct, the green is for any field that is already valid when the user comes to the form. So when the user has a defaultAccount the channel dropdown will be green whether the user is fulfilling a request or just sending money. There are other fields where this is also true: When creating a new request, if the account number is filled out it should be green. And then I believe that all fields are supposed to turn green once the user has filled them out, which is why triggering in the observer is the way to go.

app/src/main/java/com/hover/stax/views/CustomTextInputLayout.java
davkutalek: I think the real problem is that there are a lot more methods than there should be. Each of those methods is doing the exact same thing which means there is a way to simplify the code into just a couple methods.

app/src/main/java/com/hover/stax/transfers/TransferFragment.java
androidmaven: Hmmm, maybe we should ask Justin for his thoughts as well.

davkutalek:
Merging to developement. There are additional fixes pending, I will create a PR for those now

PR #37, opened 2021-02-26
Sprint3bugs fixes

androidmaven:


davkutalek (approved):
This is pretty good, grouping the message and state info makes sense, but objects need to have more substance to them. Objects should never be created because you need a convenient place to put things. Objects need to represent something, and a key indicator that you have chosen the right something is that you can move a lot of code into it that either reduces duplication or reduces other objects' visibility into the details of something they don't need to know. Reducing code duplication should always be a primary goal.

In this case I think I did lead you astray: having the errors stored in the view model increases the complexity a lot, making it hard to scale. I shouldn't have done that in the first place. The view already stores all the info for state and helper text.

There are a number of places where the state doesn't update correctly, and this is largely due to the complexity of managing everything in the viewmodel. I have a fix that is almost ready, I'll make a pull request of my own on top of this tomorrow. For now go ahead and merge this and sprint 3 into development

PR #39, opened 2021-03-02
1.1.5 release candidate

davkutalek:
There are still a few known bugs that I am attempting to wrap up today.

Instead of relying on the ViewModel to track the error states and helper text of every field on a page which doesn't scale very well, this creates a number of UI components that each have API methods that allow their state and helper text to be set directly. 

There is a abstract base class `AbstractStatefulInput` that is extended by `StaxTextInputLayout` and `StaxDropdownLayout` so that they have the same API. Then there are `ChannelDropdown`, `ActionSelect` and `ContactInput` which compose the textinput and dropdown components, presenting a similar external API.


app/src/main/java/com/hover/stax/balances/BalancesFragment.java
```diff
@@ -95,25 +99,24 @@ private void updateServices(List<Channel> channels, View view) {
...
-	private void refreshBalances(View v) {
-		if (channelDropdown.getHighlighted() != null
-					|| (balancesViewModel.getActions().getValue() !=null
-					&& balancesViewModel.getActions().getValue().size() !=0)) {
+		toggleLink(channels != null && channels.size() > 0);
+		channelDropdown.setObservers(channelDropdownViewModel, getActivity());
+	}
 
-					balancesViewModel.setAllRunning(v.getContext());
+	public void toggleLink(boolean show) {
```
davkutalek, line 107: Moved the link out of the ChannelDropdown since it is only used and one place and the dropdown no longer replaces it in the design.

app/src/main/java/com/hover/stax/channels/ChannelDropdown.java
```diff
@@ -6,90 +6,62 @@
...
+import com.hover.stax.views.AbstractStatefulInput;
 import com.hover.stax.views.StaxDropdownLayout;
-import com.hover.stax.views.StaxTextInputLayout;
 import com.squareup.picasso.Picasso;
 import com.squareup.picasso.Target;
 
 import java.util.List;
 
-public class ChannelDropdown extends TextInputLayout implements Target { //Extend this because StaxDropdownLayout already implement a view which is causing double view bug
+public class ChannelDropdown extends StaxDropdownLayout implements Target {
```
davkutalek, line 24: Made the root of StaxDropdownLayout a FrameLayout to fix this issue

app/src/main/java/com/hover/stax/channels/ChannelDropdownViewModel.java
```diff
@@ -245,42 +222,22 @@ private void logChoice(Channel channel) {
...
 	public void setChannelFromRequest(Request r) {
 		if (r != null && selectedChannels.getValue() != null && selectedChannels.getValue().size() > 0) {
 			new Thread(() -> {
 				List<Action> acts = repo.getActions(getChannelIds(selectedChannels.getValue()), r.requester_institution_id);
-				if (acts.size() <= 0) {
+				if (acts.size() <= 0)
 					acts = repo.getActions(getChannelIds(simChannels.getValue()), r.requester_institution_id);
-					if (acts.size() <= 0){
-						String fieldMessage = getApplication().getString(R.string.channel_request_fielderror, String.valueOf(r.requester_institution_id));
-						fieldState.postValue(new FieldState(FieldStateType.ERROR, fieldMessage));
```
davkutalek, line 280: Set directly from fragment instead

app/src/main/java/com/hover/stax/SplashScreenActivity.java
```diff
@@ -30,29 +40,74 @@
...
-//			finish();
-//			return;
-//		}
+	private void startSplashForegroundSequence() {
+		initSplashAnimation();
+		authenticateBiometricOrNavigateScreenAfter110sec();
+	}
 
-		Amplitude.getInstance().initialize(this, getString(R.string.amp)).enableForegroundTracking(getApplication());
+	private void startBackgroundProcesses() {
```
jumaallan, line 52: Best if initialized on the application class

app/build.gradle
```diff
@@ -95,7 +95,7 @@ dependencies {
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.3.0'
 
     // Hover SDK
-    def sdk_version = "1.7.2-stax-1.1.4-pro"
+    def sdk_version = "1.7.2-stax-1.1.5-pro"
     releaseImplementation "com.hover:android-sdk:$sdk_version"
     debugImplementation project(":hover.sdk")
     debugImplementation 'com.android.volley:volley:1.1.1'
```
jumaallan, line 101: Retrofit would be be best. Offers great features out of the box - Flows, Suspending Calls and this works well with Room DB setup on the app

app/src/main/java/com/hover/stax/SplashScreenActivity.java
jumaallan: That is initializing
* Amplitude
* Hover
* Notification Channel
* Workers - if it's work manager

app/src/main/java/com/hover/stax/SplashScreenActivity.java
```diff
@@ -30,29 +40,74 @@
...
+	private void initAmplitude() {
+		Amplitude.getInstance().initialize(this, getString(R.string.amp)).enableForegroundTracking(getApplication());
 	}
 
 	private void initHover() {
 		Hover.initialize(this);
 		Hover.setBranding(getString(R.string.app_name), R.mipmap.stax, this);
 	}
 
+	private Boolean shouldSelfDestructWhenAppVersionExpires(Boolean value) {
```
jumaallan, line 103: Not sure how this works, but you can look at Firebase Remote Configs for this - would be much simpler

app/src/main/java/com/hover/stax/SplashScreenActivity.java
```diff
@@ -73,8 +128,8 @@ private void startWorkers() {
 	private void chooseNav() {
```
jumaallan, line 128: Also look at Navigation Component. I've seen the app uses fragments, and this would make the navigation much smoother and easier to manage on the app

androidmaven: What's happening here is deciding whether to go to fulfilRequestActivity or MainActivity. Each of those activities uses fragment components. I do think this method name needs to be updated to something better.

app/src/main/java/com/hover/stax/SplashScreenActivity.java
androidmaven: Yea, one or two of these APIs doesn't work well when initialized in the Application class. An example is the Hover SDK itself.

androidmaven (approved):
Thanks for this refractor, I understand the new approach and LGTM.

I added a couple of super minors commits.
For reference.
1. On SplashScreen; Changed some ints Constants and delete splash theme: https://github.com/UseHover/Stax/compare/12da77dbe4c3...c92224689f4c

2. Renamed SplashScreen's chooseNav to navigateMainActivityOrRequestActivity: https://github.com/UseHover/Stax/compare/c92224689f4c...a81a37e8c140

3. Also added a null check for checking transfer recipient to prevent app crash: https://github.com/UseHover/Stax/compare/a81a37e8c140...d5cf51d7171f

androidmaven (changes requested):
Functional QA bug.
SessionOverLayout doesn't fire transaction fragment.

app/src/main/java/com/hover/stax/SplashScreenActivity.java
jumaallan: Makes sense - You could try to investigate why this happens on the Hover SDK side (On my Interview solution, I initialized the HoverSDK on the application class) 

But the rest would make sense if they were initialized on the Application Class.

PR #42, opened 2021-03-10
Show bounty on settings

androidmaven:
This branch is for the flow.

See bounty on the settings page.
Enter email, save offline and upload via Worker.
Show a list of bounties.
Call SDK to run action in Manual ENV.


app/src/main/java/com/hover/stax/bounty/BountyListAdapter.java
```diff
@@ -0,0 +1,109 @@
...
+	public void onBindViewHolder(@NonNull BountyListViewHolder holder, int position) {
+		BountyAction ba = bountyActionList.get(position);
+		Action a = ba.a;
+		updateStatusLayoutAndSetClickListener(holder, ba);
+		holder.content.setText(a.getFullDescription(context));
+		holder.amount.setText(a.getBountyAmountWithCurrency());
+	}
+
+	private void updateStatusLayoutAndSetClickListener(BountyListViewHolder holder, BountyAction ba) {
+		if (ba.a.bounty_is_open == 0) { //Bounty is closed and done
```
davkutalek, line 50: This is close, but the green state only shows if this user is the one that did the bounty. So the first condition should be is_open == 0 AND last transaction != null, and then add another condition for just is_open == 0 that shows the text crossed out. (see image)

Also, when there are 2 or more conditions in an if else that are the same like this you should reduce code duplication by creating a method that takes the differences as parameters and set everything there.

image

app/src/main/java/com/hover/stax/utils/StaxVolleySingleton.java
```diff
@@ -0,0 +1,166 @@
...
+import java.io.UnsupportedEncodingException;
+
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+final public class StaxVolleySingleton {
```
davkutalek, line 42: No, don't do this for something so small. Isn't there a simple way to add a header to an okhttp request?

davkutalek (changes requested):
Overall this looks good but the use of the DB/View Model/Worker for the email is serious overkill. Workers are also not meant to be used when the user is waiting, an AsyncTask would be more appropriate. 

I understand you did all this for offline use, but I think we do actually need to change it and require the user to be online. The reason is this: if we allow them to do it offline and they never go online then we don't know they are trying to submit and they don't know that we have no way of knowing they tried to submit. So then they are expecting communication and payment from us and we have no way of knowing that. 

Remember that all code is technical debt, so the less of it you can write to meet the requirements the better. Please change this to just use shared prefs and an async task to submit the email.

app/src/main/java/com/hover/stax/utils/StaxVolleySingleton.java
davkutalek: It's ok to add it to the default so that it is in every request

PR #43, opened 2021-03-12
176867312 db fix

davkutalek:
Removes Action and Sim entities and uses those from SDK which are now Room based as well.


androidmaven (approved):

davkutalek:
this was merged, just in a roundabout way

PR #44, opened 2021-03-12
Show bounty on settings

androidmaven:
Update sending user info to use AsyncTask.
Added pending states to list of bounty of actions and transaction details.
Minor copy.

There's an unresolved bug I do need help with.
- Device id returning "unknown"
- Running bounty action on my device hangs my phone. It basically doesn't initiate the session.


androidmaven:
Fixed device id showing "unknown bug" by adding basic perm check before entering bounty Activity

app/src/main/java/com/hover/stax/utils/Constants.java
```diff
@@ -36,4 +40,9 @@
 	final public static String REQUEST_LINK = "request_link";
 	final public static String FRAGMENT_DIRECT = "fragment_direct";
 	final public static String PAYMENT_LINK_SEPERATOR = "-";
+
+	final public static String BOUNTY_EMAIL = "bounty_email";
```
davkutalek, line 44: Don't use constants for things that are only used in one class. They should be private to the class they are used in. This is only for things that are used in many places, like navigation.

app/src/main/java/com/hover/stax/transactions/TransactionDetailsFragment.java
```diff
@@ -34,57 +38,115 @@ public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
...
+			if (transaction != null) {
+				setHeaderTexts(view, transaction);
+				setNormalActionPendingStateCard(view, transaction);
+			}
+		});
+	}
+
+	@SuppressLint("ResourceAsColor")
+	private void setPendingCardBgAndText(boolean isFlowDone, StaxCardView pendingCard, TextView pendingStatusText) {
+		if(isFlowDone){
```
davkutalek, line 104: Still a lot of duplicated code here, will make a video showing what I mean

davkutalek (commented):
Overall structure is pretty good. Still need to think through your objects a bit more, but you got pretty close here. I'll make a PR and video for you shortly.

PR #50, opened 2021-03-16
Minor performance improvement

androidmaven:


davkutalek:
Why are we storing an extra reference to the View, Is there a source for this? Fragments already hold their own reference (hence getVIew()) so I can't image holding an extra reference and overloading destroy functions just to call the parent is going to help performance

androidmaven:
I  had a read on: https://stackoverflow.com/questions/59503689/could-navigation-arch-component-create-a-false-positive-memory-leak . 

But I do agree if we are not referencing view as a variable, destoryView() would handle the rest. Reverted back in new commit.

davkutalek (approved):
LGTM

PR #53, opened 2021-03-23
177364518 bounty analytics tag

androidmaven:


app/src/main/java/com/hover/stax/bounties/BountyEmailFragment.java
```diff
@@ -43,12 +44,15 @@ public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceStat
...
 	@Override
 	public void onClick(View v) {
+		Amplitude.getInstance().logEvent(getString(R.string.clicked_bounty_email_continue_btn));
 		if (validates()) {
 			emailInput.setEnabled(false);
 			new BountyAsyncCaller(new WeakReference<>(getContext()), this).execute(emailInput.getText());
 			emailInput.setState(getString(R.string.bounty_uploading_email), AbstractStatefulInput.INFO);
-		} else
+		} else{
+			Amplitude.getInstance().logEvent(getString(R.string.bounty_email_err,getString(R.string.bounty_email_error)));
```
davkutalek, line 53: Probably don't need this one - this will fire anytime someone types a non-email address, which doesn't really matter to us. The offline error is important, so that one is fine.

davkutalek (approved):
Nice. Just one that I think we can remove

PR #54, opened 2021-03-24
177461250 bounty no sim state

androidmaven:
- Updated sdk to call: getPresentByHnis(String[] hnis). It was a minor addition so I just added it as a push commit to stax branch. Deployed a release 1.7.2-stax-1.2.1, which also fixes the SMS parser issue for release version

-Recent designs colors button for positive dialog btn, updated codebase to reflect this.

- No sim state

### DIFF
--- a/app/src/main/java/com/hover/stax/bounties/BountyViewModel.java
+++ b/app/src/main/java/com/hover/stax/bounties/BountyViewModel.java
@@ -10,9 +10,13 @@ import androidx.lifecycle.MutableLiveData;
 import androidx.lifecycle.Transformations;
 
 import com.hover.sdk.actions.HoverAction;
+import com.hover.sdk.sims.SimInfo;
+import com.hover.sdk.utils.Utils;
 import com.hover.stax.channels.Channel;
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/hover/stax/bounties/BountyViewModel.java
```diff
@@ -81,4 +91,13 @@ private void makeBounties(List<HoverAction> actions, List<StaxTransaction> trans
 		}
 		bountyList.setValue(bounties);
 	}
+
+	public void setSimPresentBounty(Bounty b) {
```
davkutalek, line 95: This won't actually catch SIM changes. Need to call `Hover.updateSimInfo(context);` to ensure the DB gets updated. The sim update worker only runs every 15 minutes without this manual call. See ChannelDropdown for example.

davkutalek (changes requested):
Overall works, just the one issue.

androidmaven:
Thanks for the direction, another push commit coming in a moment

PR #55, opened 2021-03-24
fixed Formatting issue in mapped flow detail page

androidmaven:


davkutalek (approved):
Looks good. The flow_done_desc string doesn't have to have CDATA this way?

androidmaven:
The  flow_done_desc is just plain text.

davkutalek:
> The flow_done_desc is just plain text.

It is being used inside the HTML call, so was just curious

androidmaven:
That's true. I'll update that with a commit. Thanks for pointing this out.

PR #56, opened 2021-03-24
177088466 filter bounties bycountry

androidmaven:


davkutalek (changes requested):
Instead of using PNGs which is going to blow up the size of our app use unicode flags. https://stackoverflow.com/a/35849652/1284291 should be exactly what you need

davkutalek:
Note for QA: test on Android 4.3 device for unicode reasons

app/src/main/java/com/hover/stax/countries/CountryAdapter.java
```diff
@@ -0,0 +1,64 @@
...
+	private static final String TAG ="CountryAdapter";
+	private String[] countryCodes;
+
+	public CountryAdapter(@NonNull  String[] countryCodes, @NonNull Context context) {
+		super(context, 0, countryCodes);
+		this.countryCodes = countryCodes;
+
+	}
+
+	@SuppressLint("ViewHolder")
```
davkutalek, line 27: If you instead just inflate the view only if it is null then it won't give you this ViewHolder error and you'll get the same performance as wit viewholder since it will reuse the views. You're probably better off not using a viewholder here anyway since it basically adds more boilerplate code. This is all fine as-is, this is just a note

app/src/main/java/com/hover/stax/countries/CountryDropdown.java
```diff
@@ -0,0 +1,106 @@
...
+
+public class CountryDropdown extends StaxDropdownLayout {
+	private static final String TAG = "CountryDropdown";
+	private CountryAdapter.SelectListener selectListener;
+
+	public CountryDropdown(@NonNull Context context, @Nullable AttributeSet attrs) {
+		super(context, attrs);
+	}
+
+	public void updateChoicesByHoverActions(List<HoverAction> actionList) {
```
davkutalek, line 32: Remove these 2 unused functions

app/src/main/java/com/hover/stax/utils/UIHelper.java
```diff
@@ -123,4 +123,17 @@ public static void loadPicasso(String url, int size, Target target) {
...
 	}
+
+	public static void loadPicasso(int resId, int size, Target target) {
+		Picasso.get()
+				.load(resId)
+				.config(Bitmap.Config.RGB_565)
+				.networkPolicy(NetworkPolicy.OFFLINE)
+				.resize(size,size).into(target);
+	}
+	public static String countryCodeToEmoji(String countryCode) {
```
davkutalek, line 134: make this a static function in CountryDropdown, no need for it to be here

app/src/main/java/com/hover/stax/bounties/BountyListFragment.java
```diff
@@ -48,17 +57,28 @@ private void startObservers() {
...
+		bountyViewModel.filterChannels(countryCode).observe(getViewLifecycleOwner(), channels -> {
+			if (channels != null && channels.size() > 0 && bountyViewModel.getBounties().getValue() != null && bountyViewModel.getBounties().getValue().size() > 0) {
+				createList(channels, bountyViewModel.getBounties().getValue(), false);
+			}
 		});
 	}
 
-	private void createList(List<Channel> channels, List<Bounty> bounties) {
+	private void createList(List<Channel> channels, List<Bounty> bounties, boolean shouldUpdateCountryList) {
+		if(shouldUpdateCountryList) countryDropdown.updateChoicesByChannels(channels);
```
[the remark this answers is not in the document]
davkutalek: `updateChoicesByChannels` only needs to be called when the channels update anyway, doesn't need to be called when bounties update.

app/src/main/java/com/hover/stax/bounties/BountyListFragment.java
```diff
@@ -80,4 +100,9 @@ public void bountyDetail(Bounty b) {
 			})
 			.showIt();
 	}
+
+	@Override
+	public void countrySelect(String countryCode) {
```
davkutalek, line 105: This doesn't need to call through to another function. filterBounties is only called from here, so just put the logic in here

app/src/main/java/com/hover/stax/bounties/BountyListFragment.java
```diff
@@ -48,17 +57,28 @@ private void startObservers() {
...
-			if (channels != null && channels.size() > 0 && bountyViewModel.getBounties().getValue() != null && bountyViewModel.getBounties().getValue().size() > 0)
-				createList(channels, bountyViewModel.getBounties().getValue());
+			if (channels != null && channels.size() > 0 && bountyViewModel.getBounties().getValue() != null && bountyViewModel.getBounties().getValue().size() > 0) {
+				createList(channels, bountyViewModel.getBounties().getValue(), true);
+			}
+		});
+	}
+	private void filterBounties(String countryCode) {
+		bountyViewModel.filterChannels(countryCode).observe(getViewLifecycleOwner(), channels -> {
+			if (channels != null && channels.size() > 0 && bountyViewModel.getBounties().getValue() != null && bountyViewModel.getBounties().getValue().size() > 0) {
```
davkutalek, line 74: Since we are repeating this same null check everywhere, move it into the `createList` function. Also rename createList, bad naming on my part.

davkutalek (approved):
Looks great! Asking for a few minor changes, but nothing serious

app/src/main/java/com/hover/stax/countries/CountryAdapter.java
androidmaven: Thanks!

PR #57, opened 2021-03-24
177479494 push notification

androidmaven:


app/src/main/java/com/hover/stax/ApplicationInstance.java
```diff
@@ -23,5 +24,7 @@ public void onCreate() {
 		Lingver.init(this, Locale.getDefault());
 
 		FirebaseApp.initializeApp(this);
+		FirebaseMessaging.getInstance().setAutoInitEnabled(true);
```
davkutalek, line 27: default is true

app/src/main/java/com/hover/stax/ApplicationInstance.java
```diff
@@ -23,5 +24,7 @@ public void onCreate() {
...
+		FirebaseMessaging.getInstance().setDeliveryMetricsExportToBigQuery(true);
```
davkutalek, line 28: Do we need this?

app/src/main/java/com/hover/stax/channels/ChannelDropdown.java
```diff
@@ -94,6 +95,11 @@ private void onSelect(Channel c) {
 		setDropdownValue(c);
 		if (highlightListener != null) { highlightListener.highlightChannel(c); }
 		highlightedChannel = c;
+		if(c!=null) {
+			FirebaseMessaging.getInstance().subscribeToTopic(
```
davkutalek, line 99: This should be in the logChoice function in the ChannelDropdownViewModel

app/src/main/java/com/hover/stax/pushNotification/FirebaseMessageReceiver.java
```diff
@@ -0,0 +1,74 @@
...
+
+	@Override
+	public void
+	onMessageReceived(RemoteMessage remoteMessage) {
+		if (remoteMessage.getNotification() != null) {
+			showNotification(remoteMessage.getNotification().getTitle(), remoteMessage.getNotification().getBody());
+		}
+	}
+
+	private RemoteViews getCustomDesign(String title, String message) {
```
davkutalek, line 36: Why are we making our own design rather than using the default which already contains a title, message and icon?

davkutalek (changes requested):
Looks good, only big thing is the notification layout, not convinced it needs to be custom. Can you put links to those loom videos you made in the pivotal story?

app/src/main/java/com/hover/stax/ApplicationInstance.java
androidmaven: Export metrics to big query is also available on Firebase on the fly might be useful especially when our user base grow exponential.

app/src/main/java/com/hover/stax/pushNotification/FirebaseMessageReceiver.java
androidmaven: No data to support this: but personally, I tend to engage with notifications that have a custom layout, cos they are just different from regular ones; eyes tend to focus on the different ones.

I think this might help our engagement, but I'd put this in Pivotal for Justin.

app/src/main/java/com/hover/stax/channels/ChannelDropdown.java
androidmaven: Yup, will update, thanks.

app/src/main/java/com/hover/stax/channels/ChannelDropdownViewModel.java
```diff
@@ -162,6 +162,7 @@ protected void setActiveChannelIfNull(List<Channel> channels) {
 	}
 
 	private void setActiveChannel(Channel channel) {
+		FirebaseMessaging.getInstance().subscribeToTopic(getApplication().getString(R.string.firebase_topic_institution, String.valueOf(channel.institutionId)));
```
davkutalek, line 165: Still not right. There is a function called logChoice that already subscribes to firebase channels

app/src/main/java/com/hover/stax/ApplicationInstance.java
davkutalek: Don't enable things until they are needed. We already have to many data and analytics tools going on

davkutalek (approved):
LGTM. Minor changes, will merge and correct in a moment.

app/src/main/java/com/hover/stax/channels/ChannelDropdownViewModel.java
davkutalek: Actually we don't want to subscribe to Institutions anyway. Institutions don't have a country, so it doesn't really help us. By subscribing users to a channel we can always connect to their institution if we want to

PR #60, opened 2021-03-31
177562755 stax onboarding

androidmaven:


davkutalek (changes requested):
What is here is good, but seems like serious overkill. Why do we need to use individual fragments and deal with all the animating ourselves? This implementation is much simpler: https://www.geeksforgeeks.org/image-slider-in-android-using-viewpager/ then just change pages by using `viewPager.setCurrentItem()` on a delayed handler.

I've included this is in the QA build, but I would like to get the simplified version in by release wed if possible,

androidmaven:
There are 4 reasons why I chose the current approach over  https://www.geeksforgeeks.org/image-slider-in-android-using-viewpager/

1. Each slide needs to be a layout; because the images provided by Justin were without the texts. 
2. Each slide layouts had to be in its own individual XML, because the positioning of the texts is different.
3. If the texts were to come merged with the image, so we use images, not layouts. This might cause serious scaling issues for some screen sizes. Texts might go off the screen etc.
4. I initially tried using a delay timer, but there was no way to handle ease animation from one slide to another.

davkutalek (approved):
This is actually great as is then, thanks

PR #68, opened 2021-04-19
177626303 smarter contact matching

androidmaven:


app/src/main/java/com/hover/stax/transfers/TransferActivity.java
```diff
@@ -66,16 +66,31 @@ private void createFromSchedule(int schedule_id) {
...
+	private void observeRequest() {
+		AlertDialog dialog = new StaxDialog(this).setDialogMessage(R.string.loading_link_dialoghead).showIt();
 		transferViewModel.getRequest().observe(this, request -> {
 			Log.e(TAG, "maybe viewing request");
 			if (request == null) return;
+
 			Log.e(TAG, "viewing request " + request);
-			if (dialog != null) dialog.dismiss();
+			if (dialog != null) {
+				transferViewModel.setRecipientSmartly(request, channelDropdownViewModel.getActiveChannel().getValue());
```
davkutalek, line 83: setRecipientSmartly should be called from the fragment. There is already an observer there for request, so this should go in the load function which is called when the request is loaded from the DB. Same to the channel observer, there is already one there, use it instead.

davkutalek (changes requested):
Looks good! Just one change please

alexosugo (approved):

alexosugo (approved):

PR #74, opened 2021-04-23
177665390 viewbinding

alexosugo:
- Updates the app to use ViewBinding library.
- Refactors ArrayAdapters to use ViewHolder pattern.
- Flattens some layouts for easier accessibility by the ViewBinding library.


app/src/main/java/com/hover/stax/balances/BalanceAdapter.java
```diff
@@ -43,32 +43,34 @@ public void showBalance(boolean show) {
...
 	}
 
 	@RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
 	@Override
 	public void onBindViewHolder(@NonNull BalanceViewHolder holder, int position) {
 		Channel channel = channels.get(position);
 
-		holder.channelName.setText(channel.name);
-		holder.channelId.setText(Integer.toString(channel.id));
+		holder.binding.balanceChannel.setText(channel.name);
```
androidmaven, line 55: What of: ...balanceChannelName being a better naming?

alexosugo: Agreed. Change effected.

PR #83, opened 2021-05-04
fix to radio button always stuck at showing -myself- even when change…

androidmaven:
…d to someone else

### DIFF
--- a/app/src/main/java/com/hover/stax/transfers/TransferFragment.java
+++ b/app/src/main/java/com/hover/stax/transfers/TransferFragment.java
@@ -107,7 +107,9 @@ public class TransferFragment extends AbstractFormFragment implements ActionSele
 			if (transferViewModel.getContact().getValue() != null)
 				contactInput.setSelected(transferViewModel.getContact().getValue());
 		});
-		transferViewModel.getContact().observe(getViewLifecycleOwner(), contact -> recipientValue.setContact(contact));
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/hover/stax/transfers/TransferFragment.java
```diff
@@ -170,8 +170,18 @@ private void setRecipientHint(HoverAction action) {
...
 			recipientValue.setContent(getString(R.string.self_choice), "");
-		else
+		else {
+			addRecipientsValue();
 			contactInput.setHint(action.getRequiredParams().contains(HoverAction.ACCOUNT_KEY) ? getString(R.string.recipientacct_label) : getString(R.string.recipientphone_label));
+		}
+	}
+
+	private void addRecipientsValue() {
+		StaxContact contact = transferViewModel.getContact().getValue();
```
alexosugo, line 180: Ideal usage of ViewModels is to let the observer handle the data change and update the UI. We shouldn't fetch the value directly from the MutableLiveData object.

alexosugo (changes requested):

app/src/main/java/com/hover/stax/transfers/TransferFragment.java
androidmaven: Very true, thanks for this correction.

alexosugo (approved):

PR #84, opened 2021-05-04
177947406 new stax notification groups

androidmaven:


app/src/main/java/com/hover/stax/requests/NewRequestFragment.java
```diff
@@ -193,10 +194,16 @@ public void onTextChanged(CharSequence s, int start, int before, int count) {
 
 	private void fabClicked(View v) {
 		requestViewModel.removeInvalidRequestees();
-		if (requestViewModel.getIsEditing().getValue() && validates())
+		if (requestViewModel.getIsEditing().getValue() && validates()) {
```
alexosugo, line 197: Can the same effect be achieved by observing the object instead?

alexosugo (changes requested):

app/src/main/java/com/hover/stax/requests/NewRequestFragment.java
androidmaven: This function is called and needed only when the fab is clicked. 

How do you suggest we observe the object?

alexosugo: The observer can be registered when the fab is clicked.

androidmaven: Hmm, then why do we need to register an observer in the view model to call a function that's needed and available on the activity class; when we could just call the function directly from the activity class?

e.g `validates()`

alexosugo: We don't need to register an observer for the `validates()` function. I pointed this out because the static analysis shows that access to this variable may be null. For this use case, it can be used as is, as long as we set a default state for this boolean. Could you point me to where the value for `setEditing()` is called?

alexosugo (approved):

PR #107, opened 2021-05-19
sets up Sentry for crash reporting

alexosugo:


app/src/main/res/values/strings.xml
```diff
@@ -361,6 +361,6 @@
 	<string name="language">Language</string>
 	<string name="bounty">Bounty</string>
 
-
+	<string name="sentry_dsn">https://ce9bf386ab104ece903cd33fc563c241@o44209.ingest.sentry.io/5477330</string>
```
davkutalek, line 364: Put this at the top of the file and don't forget `translatable="false"`

app/src/main/java/com/hover/stax/ApplicationInstance.java
```diff
@@ -45,6 +51,29 @@ private void setLogger() {
             Timber.uprootAll();
         }
     }
+
+    private void initSentry(){
```
davkutalek, line 55: Do we have to use manual initialization for Stax?

alexosugo: From my tests, it seems the `SDK` configurations override Stax's and the events aren't recorded.

davkutalek: Cool, was just curious. It does work this way?

PR #109, opened 2021-05-20
178131341 remote trigger self destruct

androidmaven:


app/src/main/java/com/hover/stax/navigation/AbstractNavigationActivity.java
```diff
@@ -82,15 +82,16 @@ protected NavController getNavController() {
     }
 
     public void checkPermissionsAndNavigate(int toWhere) {
```
davkutalek, line 84: This looks wrong, why is permissions check commented?

davkutalek:
Don't forget to put bubble showcase back before merging

app/src/main/java/com/hover/stax/navigation/AbstractNavigationActivity.java
androidmaven: Thanks, was an oversight to refractor back.

alexosugo:
Closing this as it has been merged ti development.

PR #116, opened 2021-06-02
completed story for: As a user I want balances to be on the home screen and to see account placeholders when I first open the app.

androidmaven:
- Refactored codebase.


davkutalek (approved):
LGTM

PR #117, opened 2021-06-02
completed story for: As a user, I want to see my requests in my account's details page

androidmaven:


app/src/main/java/com/hover/stax/channels/ChannelDetailFragment.java
```diff
@@ -62,10 +84,28 @@ public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceStat
         viewModel.getFeesThisYear().observe(getViewLifecycleOwner(), sum -> {
             binding.detailsFees.setText(Utils.formatAmount(sum != null ? sum : 0));
         });
+    }
+
+    private void setUpFuture(Channel channel) {
+        futureViewModel.getScheduledByChannel(channel.id).observe(getViewLifecycleOwner(), schedules -> {
```
alexosugo, line 90: Since this event will be fired each time there's a change in the dataset, how about the `RecyclerView` is initialised outside this block then change the adapter or notifyDataSetChanged when there are updates?

androidmaven: Thanks for this, codebase updated

davkutalek: There is still a small typo here - the class variable is not being set. Fixed in upcoming merge

alexosugo (approved):

PR #119, opened 2021-06-07
178200976 balances animation

alexosugo:
- Introduces animations for the balances fragment; tap to hide or show. At the moment, the whole `RecyclerView` is resized to achieve the effect. Individual line items can also be animated but it'll take more time and is more prone to errors. One constraint that was taken into account is that backgrounds are overridden in the `MotionScene`. Wrapping the `RecyclerView` in a `CardView` mitigates against this since we can set the background color. However, you cannot set a corner radius for a specific side. It's all or nothing.


davkutalek (approved):
LGTM. Only minor issue is that the accounts are hidden by default even when they are only placeholders. They should show so long as they are not real accounts.

I would have expected that a RecyclerView would cause a headache and it would be easier to just append layouts to eg a LinearLayout manually, but if this works then awesome.

PR #123, opened 2021-06-11
completed story for- As a user I want to give app reviews without getting annoyed

androidmaven:


app/src/main/java/com/hover/stax/home/MainActivity.java
```diff
@@ -89,23 +98,48 @@ else if(deepLinkRoute.contains(getString(R.string.deeplink_settings))) {
...
+                launchStaxReview();
             }
         }
     }
+    private void observeTransactionsForAppReview() {
+        transactionHistoryViewModel.getStaxTransactionsForAppReview().observe(this, staxTransactions -> {
+                if(qualifiesForAppReview(staxTransactions)) launchRatingAndReviewDialog();
+        });
+    }
+    private boolean qualifiesForAppReview(List<StaxTransaction> staxTransactions) {
```
alexosugo, line 111: Let's have this happen in the `ViewModel` and only post the result to the fragment. It's better for performance.

alexosugo (changes requested):

app/src/main/java/com/hover/stax/home/MainActivity.java
androidmaven: Thanks for this, refractor done.

alexosugo (approved):

PR #126, opened 2021-06-14
178200987 new bubble flow

androidmaven:


alexosugo (changes requested):
Animation works great. The bubble shows up as expected. Two things however:
- Is the bubble show case supposed to appear every time I go to the `Home` fragment when I haven't added an account?
- The text on the first line on the showcase appears cut off at the end. See the attached image. 
Image 2021-06-15 at 09 12 12

androidmaven:
@alexosugo From my understanding from the UX. Yes, the "first" bubble should show if you don't have any account. But once you have 1, the "second" bubble should only show once.

 - Thanks for the second UI bug catch, a fix is enroute.

PR #129, opened 2021-06-14
178520182 get paid navigation bugfixes

androidmaven:
Fix for: Back from send money or request always leaves Get paid highlighted


alexosugo (changes requested):
Issue still persists. Navigating back still leaves `Get Paid` highlighted. Perhaps we could monitor the destination changes for the top level menu items and highlight the respective ones.

alexosugo (approved):

PR #147, opened 2021-06-28
v1.4.1

alexosugo:


app/src/main/java/com/hover/stax/bounties/BountyAsyncCaller.java
```diff
@@ -14,18 +13,21 @@
...
 
 import okhttp3.MediaType;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.RequestBody;
 import okhttp3.Response;
+import timber.log.Timber;
 
-class BountyAsyncCaller extends AsyncTask<String, Void, Integer> {
+class BountyAsyncCaller extends AsyncTask<String, Void, Map<Integer, String>> {
```
davkutalek, line 26: This should be replaced with modern implementation

davkutalek (approved):
I don't see any issues with anything, but given the size of the diff and changes in tabbing as well as java > kt there is plenty that I didn't look at in detail. Is there anything you want me to look at in particular?

.github/workflows/prod-release.yml
```diff
@@ -0,0 +1,117 @@
...
+        with:
+          releaseDirectory: app/build/outputs/bundle/release
+          signingKeyBase64: ${{ secrets.SIGNING_KEY }}
+          alias: ${{ secrets.ALIAS }}
+          keyStorePassword: ${{ secrets.KEY_STORE_PASSWORD }}
+          keyPassword: ${{ secrets.KEY_PASSWORD }}
+
+
+      # Create release with tag on Github
+      - name: Create Release
```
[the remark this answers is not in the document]
alexosugo: I'll need file to edit this file to just create the release tag and not the release build. That way it doesn't break the current release structure we have.

app/src/main/java/com/hover/stax/bounties/BountyAsyncCaller.java
alexosugo: Yes. I'm gonna add this to Pivotal.

PR #158, opened 2021-07-13
178726673 completed story for in app banners

androidmaven:
Logic description here: https://www.notion.so/usehover/In-app-messaging-89385fdfe7a64fa4a07aeae91b90a528#f192605153ce4fd4baa535ff78327b49

1. One thing not yet added on notion which I'll sync with Peace on this- 28 days recurring if permission not accepted
2. For faster functional QA, I'd suggest you comment out: BannerUtils, line 131
3. DateUtils class was converted to Kotlin; main code difference is in from line 135 to 160


app/src/main/java/com/hover/stax/database/DatabaseRepo.kt
```diff
@@ -0,0 +1,318 @@
...
+    fun delete(request: Request?) {
+        AppDatabase.databaseWriteExecutor.execute { requestDao.delete(request) }
+    }
+
+    companion object {
+        private const val TAG = "DatabaseRepo"
+    }
+
+    init {
+        val db = AppDatabase.getInstance(application)
```
alexosugo, line 306: The app and sdk databases can be passed as constructor arguments in the class declaration. You can then modify the DatabaseRepo instantiation in the `module.kt` file to make them injectable.

app/src/main/java/com/hover/stax/inapp_banner/BannerViewModel.kt
```diff
@@ -0,0 +1,27 @@
...
+    private val bannerUtils = BannerUtils(getApplication())
+
+    init {
+            viewModelScope.launch {
+                val hasTransactionLastMonth: Boolean = repo.hasTransactionLastMonth()
+                qualifiedBannerLiveData.postValue(bannerUtils.getQualifiedBanner(hasTransactionLastMonth))
+            }
+    }
+
+    fun qualifiedBanner(): LiveData<Banner> = qualifiedBannerLiveData
```
alexosugo, line 22: We can directly access the `qualifiedBannerLiveData` object so this is necessary.

app/src/main/java/com/hover/stax/inapp_banner/BannerViewModel.kt
```diff
@@ -0,0 +1,27 @@
...
+import androidx.lifecycle.AndroidViewModel
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.viewModelScope
+import com.hover.stax.database.DatabaseRepo
+import kotlinx.coroutines.launch
+
+class BannerViewModel(application: Application, repo: DatabaseRepo) : AndroidViewModel(application) {
+    private val qualifiedBannerLiveData: MutableLiveData<Banner> = MutableLiveData<Banner>()
+    private val bannerUtils = BannerUtils(getApplication())
```
alexosugo, line 13: Use the injected application instance in the constructor.

app/src/main/java/com/hover/stax/utils/DateUtils.kt
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    fun humanFriendlyTime(context: Context, diffMillis: Long): String {
+        val seconds = (Math.abs(diffMillis) / 1000).toDouble()
+        val minutes = seconds / 60
+        val hours = minutes / 60
+        val days = hours / 24
+        val years = days / 365
+        val time: String
+        time = if (seconds < 45) {
```
alexosugo, line 52: Can relative timestamp work for this?

alexosugo (changes requested):

app/src/main/java/com/hover/stax/inapp_banner/BannerViewModel.kt
androidmaven: Directly accessing it counters a clean approach to getting results from LiveData. 

- Variables should be separate from its getter.

I understand kotlin does something similar for data classes, but it this case, kotlin under the hood generates getters and setter methods.

app/src/main/java/com/hover/stax/database/DatabaseRepo.kt
androidmaven: I tried this, but this will require refactoring all ViewModels, which was causing more errors, and I'm unable to test because certain bugs in this branch aren't resolved here, instead resolve in the base branch.

Easier to do the refractoring on development.

app/src/main/java/com/hover/stax/inapp_banner/BannerViewModel.kt
androidmaven: Thanks, done this

app/src/main/java/com/hover/stax/utils/DateUtils.kt
androidmaven: This isn't a new code. 
However, the customization was made so that we can extend more Date functions easily.

alexosugo (approved):

PR #159, opened 2021-07-13
Added library activity and other parts to create an index of USSD cod…

davkutalek:
…es that can be dialed directly


davkutalek:
@alexosugo lets include this in the next release

alexosugo (approved):

PR #160, opened 2021-07-14
Multivariate Testing

alexosugo:


app/src/main/java/com/hover/stax/requests/Request.java
```diff
@@ -123,7 +122,7 @@ String generateWhatsappRecipientString(List<StaxContact> contacts, Channel c) {
 
     public String generateMessage(Context c) {
         String amountString = amount != null ? c.getString(R.string.sms_amount_detail, Utils.formatAmount(amount)) : "";
-        String noteString = note != null && !TextUtils.isEmpty(note) ? c.getString(R.string.sms_note_detail, note) : "";
+        String noteString = note != null && !TextUtils.isEmpty(note) ? c.getString(R.string.sms_note_detail, note) : " ";
```
davkutalek, line 125: I don't think we want an empty space when the note is empty?

app/src/main/java/com/hover/stax/transfers/TransactionType.kt
```diff
@@ -0,0 +1,13 @@
+package com.hover.stax.transfers
+
+import timber.log.Timber
+import kotlin.properties.Delegates
+
+class TransactionType {
```
davkutalek, line 6: Could you explain why this is like this?

davkutalek (commented):
@alexosugo finally finished this review. Looks great, I really don't have anything to add. One question in comments. I'll review the more recent code today as well

app/src/main/java/com/hover/stax/requests/Request.java
alexosugo: Yeah. I shouldn't be empty.

app/src/main/java/com/hover/stax/transfers/TransactionType.kt
alexosugo: The issue was with how the ViewModels are initialised from the `Koin` module. If we access the `ViewModel` normally where an activity + fragment or 2 fragments share one, then a new instance of the `ViewModel` will be created for each fragment. Therefore, all variables will be reset to their original values making passing info and state difficult. To solve this, we use access the `ViewModel` in either classes as a `SharedViewModel`. You can see this in action in the `BalancesFragment`.

If two fragments share a `ViewModel` but are not hosted in the same activity, then the method above fails. The only way I could find to send the transaction type was to use a delegates.

PR #165, opened 2021-07-15
178726690 faq

androidmaven:


app/src/main/java/com/hover/stax/faq/FAQ.kt
```diff
@@ -0,0 +1,40 @@
...
+                return@addOnSuccessListener
+            }.addOnFailureListener { exception ->
+                Timber.e(exception)
+                return@addOnFailureListener
+            }
+
+    return liveData
+}
+
+class FaqViewModel(application: Application) : AndroidViewModel(application) {
```
alexosugo, line 38: Since we're not using the application instance, this can be simplified to use the `ViewModel` class with no params.

app/src/main/java/com/hover/stax/faq/FAQAdapter.kt
```diff
@@ -0,0 +1,46 @@
...
+
+import android.view.LayoutInflater
+import android.view.ViewGroup
+import androidx.recyclerview.widget.RecyclerView
+import com.hover.stax.databinding.FaqItemsBinding
+
+class FAQAdapter(private val faqs: List<FAQ>, private val selectListener: SelectListener):
+        RecyclerView.Adapter<FAQAdapter.Companion.FAQViewHolder>() {
+
+    companion object {
```
alexosugo, line 11: The viewholder can be defined as an inner class to allow for access to it and the listener. Then you don't have to wrap it in a companion object.

app/src/main/java/com/hover/stax/faq/FaqFragment.kt
```diff
@@ -0,0 +1,93 @@
...
+import androidx.fragment.app.Fragment
+import com.hover.stax.R
+import com.hover.stax.databinding.FragmentFaqBinding
+import com.hover.stax.utils.UIHelper
+import com.hover.stax.utils.Utils
+import org.koin.androidx.viewmodel.ext.android.getViewModel
+
+class FaqFragment : Fragment(), FAQAdapter.Companion.SelectListener {
+
+    lateinit var binding: FragmentFaqBinding
```
alexosugo, line 18: Fragments outlive their views so we need to clean up binding references when the view is destroyed. This should be initialized as a nullable variable.

alexosugo (changes requested):

alexosugo (approved):

PR #166, opened 2021-07-16
178595281 improved filter bounties

androidmaven:
- added loading state for bounty filter by country
- added: hide bounties that are closed and not done by the user


alexosugo (commented):
The loading and filters work great. However, would it be prudent to hide channel cards in instances where there are no open bounties? Check the case of Stanbic in the image below. 

Bounties

androidmaven:
Thanks for this @alexosugo , I've effected the update request. As well as improved how the closed bounty are made invisible.

alexosugo (approved):

PR #206, opened 2021-07-30
178972396 bounty email bugfix

androidmaven:
- Converted bounty email background work to use coroutines instead of AsyncTask
- Updated makeBounties() to work in the background using coroutines because we now have long list of bounties.
- BountyEmailFragment and BountyViewModel converted to kotlin class, since it's required to make use of coroutines.


app/src/main/java/com/hover/stax/bounties/BountyEmailFragment.kt
```diff
@@ -0,0 +1,114 @@
...
+            } else {
+                emailInput!!.setState(getString(R.string.bounty_email_error), AbstractStatefulInput.ERROR)
+            }
+        } else {
+            showOfflineDialog()
+        }
+    }
+
+    private fun showOfflineDialog() {
+        StaxDialog(requireActivity())
```
alexosugo, line 60: This should be assigned to the dialog variable declared above.

app/src/main/java/com/hover/stax/bounties/BountyEmailFragment.kt
```diff
@@ -0,0 +1,114 @@
...
+                .setDialogTitle(R.string.internet_required)
+                .setDialogMessage(R.string.internet_required_bounty_desc)
+                .setPosButton(R.string.btn_ok, null)
+                .makeSticky()
+                .showIt()
+    }
+
+    private fun showEdgeCaseErrorDialog() {
+        StaxDialog(requireActivity())
```
alexosugo, line 69: Same goes for this one.

app/src/main/java/com/hover/stax/bounties/BountyEmailFragment.kt
```diff
@@ -0,0 +1,114 @@
...
+
+    private fun saveAndContinue() {
+        logAnalyticsEvent(getString(R.string.bounty_email_success), requireContext())
+        saveString(BountyActivity.EMAIL_KEY, emailInput!!.text, requireActivity())
+        NavHostFragment.findNavController(this).navigate(R.id.action_bountyEmailFragment_to_bountyListFragment)
+    }
+
+    override fun onDestroyView() {
+        super.onDestroyView()
+        if (dialog != null && dialog!!.isShowing) dialog!!.dismiss()
```
alexosugo, line 107: The dialog is always null since as it's not assigned to anything.

app/src/main/java/com/hover/stax/bounties/BountyViewModel.kt
```diff
@@ -0,0 +1,144 @@
...
+            val result = BountyEmailNetworking(getApplication()).uploadBountyUser(email)
+            bountyEmailLiveData.postValue(result)
+        }
+    }
+
+    private fun loadSims() {
+        if (sims == null) {
+            sims = MutableLiveData()
+        }
+        Thread { sims!!.postValue(repo.presentSims) }.start()
```
alexosugo, line 49: Since you're already using coroutines, you can replace the uses of `Thread` here.

alexosugo (changes requested):

app/src/main/java/com/hover/stax/bounties/BountyEmailFragment.kt
androidmaven: Thanks, this is now fixed in recent push

app/src/main/java/com/hover/stax/bounties/BountyEmailFragment.kt
androidmaven: Thanks, this is now fixed in recent push

app/src/main/java/com/hover/stax/bounties/BountyViewModel.kt
androidmaven: Thanks, this is now fixed in recent push

alexosugo (approved):

PR #270, opened 2021-08-19
179196298 transaction details: to properly show succedded, pending and failed status ard

androidmaven:
- Made transaction status its own component.


alexosugo (changes requested):
Code-wise looks good. The is a padding issue, however. Please check the image below. 

WhatsApp Image 2021-08-23 at 10 49 02

app/src/main/java/com/hover/stax/views/StaxTransactionStatusView.kt
```diff
@@ -0,0 +1,78 @@
...
+        when (transaction.status) {
+            Transaction.SUCCEEDED -> setSuccessView(transaction.isRecorded)
+            Transaction.PENDING -> setPendingView(transaction.isRecorded)
+            Transaction.FAILED -> setFailedView(transaction.isRecorded)
+        }
+        fillFromAttrs()
+    }
+
+    @SuppressLint("ResourceAsColor")
+    private fun setSuccessView(isBounty: Boolean) {
```
davkutalek, line 46: These functions are not DRY. Refactor so that they are one function with the setting of each parameter passed in.

davkutalek (approved):
Code looks good, just one minor cleanup I noticed.

We may need to chat about what to show for the failure case, the message "confirming that the transaction has failed" seems weird to me and we should indicate where they can find out why. Not a blocker for merge.

app/src/main/java/com/hover/stax/views/StaxTransactionStatusView.kt
androidmaven: Thanks, code is now updated. Kindly review

davkutalek (changes requested):
No, this is worse. Now everything is interdependent. I'll make a PR with what I mean tomorrow

davkutalek:
@androidmaven the TransactionDetailsFragment and PopupTransactionDetailsFragment need to be merged, that is too much duplication

PR #290, opened 2021-08-24
Refactored StaxTransactionStatusView to make it DRYer

davkutalek:
This way all the boiler plate (`with`, `when`, `binding`) stuff is not duplicated in each method and the presentation logic (Html.fromHtml, the binding ids) are separate from the resource selection logic (transaction data). 

Since this is already getting a bit complicated, the next step is to separate out the getX functions into a wrapper object. I'll continue this PR with that change in a moment.


davkutalek:
Second commit uses the [decorator pattern](https://refactoring.guru/design-patterns/decorator) to make it so that the View file does not need to know the internals about how the resources are looked up from the db data. 

I think from here you could actually get rid of the StaxTransactionStatusView by moving the flat option and updateState function to StaxCard which might be simpler.

alexosugo:
Here's a thought. Would [sealed classes](https://www.section.io/engineering-education/sealed-classes-in-kotlin/) have worked for this?

androidmaven:
Thanks! Checking through both

davkutalek:
@alexosugo not quite, the sealed class involves inheritance and is meant to allow for multiple different subclasses without allowing anyone to create arbitrary new subclasses. The advantage of the decorator pattern is that it is a case of composition rather than inheritance, which is usually preferable. Looks like in kotlin (delegation)[https://kotlinlang.org/docs/delegation.html] is designed for this case. I was a bit lazy, just did the simplest possible thing.

PR #303, opened 2021-08-31
179413415 new permission design

androidmaven:


app/src/main/java/com/hover/stax/views/StaxDialog.java
```diff
@@ -40,9 +41,17 @@ public StaxDialog setDialogTitle(int title) {
         return this;
     }
 
+    public StaxDialog setCustomView(int viewRes) {
```
alexosugo, line 44: Here, we can update the constructor to take a layout as a parameter and have one central point of inflating the layout.

alexosugo (changes requested):

app/src/main/java/com/hover/stax/views/StaxDialog.java
androidmaven: Thanks for this, it's updated in the recent push.

alexosugo (approved):

PR #314, opened 2021-09-02
179267572 retry button

androidmaven:
- Retry button on failed transactions
- When clicked thrice, should show contact support button.

- To do functional QA, you might need to change `TransactionDetailsFragment.kt, line 158 ( else if(transaction.status == Transaction.FAILED) ) to either SUCCEEDED OR FAILED,` for easier stating.


app/src/main/java/com/hover/stax/ApplicationInstance.kt
```diff
@@ -80,6 +81,10 @@ class ApplicationInstance : Application() {
         AppsFlyerLib.getInstance().init(getString(R.string.appsflyer_key), conversionListener, this)
     }
 
+    companion object {
```
davkutalek, line 84: I don't love this, in general try to avoid adding to applicationinstance. 

@alexosugo what do you think, is there a better approach? My first thought is to use shared prefs

app/src/main/java/com/hover/stax/utils/Constants.kt
```diff
@@ -28,6 +28,7 @@ object Constants {
     const val LANGUAGE_CHECK = "Language"
     const val PERM_ACTIVITY = "com.hover.stax.permissions.PermissionsActivity"
     const val TRANSFERRED = "TRANSFERRED"
```
davkutalek, line 30: Constant name and value should be "TRANSFER" so as not to imply that the transfer definitely happened

davkutalek (approved):
Looks good, minor questions.

@androidmaven try not to put fixes together with new features, makes it hard to ensure stable build or make sure that fixes get included

app/src/main/java/com/hover/stax/utils/Constants.kt
androidmaven: Actually, the instances where this constant is being used implies that the transfer definitely happened

app/src/main/java/com/hover/stax/ApplicationInstance.kt
androidmaven: Merging for now, will open a chore on pivotal tracker to make this better

PR #316, opened 2021-09-02
completed story for: As a user I want obvious transaction session tim…

androidmaven:
- Compare length of session messages extra returned to transaction receiver. If it is 0 or much shorter than length of action's steps then mark the transaction as failed


app/src/main/java/com/hover/stax/database/DatabaseRepo.kt
```diff
@@ -169,7 +176,13 @@ class DatabaseRepo(db: AppDatabase, sdkDb: HoverRoomDatabase) {
         }
     }
 
-
+    private fun shouldFlagFailure(transaction: StaxTransaction, c: Context?) : Boolean {
```
davkutalek, line 179: This name is confusing, should be something like `isSessionIncomplete` and it should be a method on the transaction model. You also don't need to use the UssdCallResponse object or creation method at all, you can just use `Hover.getTransaction(transaction.uuid, c).ussdMessages`

app/src/main/java/com/hover/stax/database/DatabaseRepo.kt
```diff
@@ -150,17 +152,22 @@ class DatabaseRepo(db: AppDatabase, sdkDb: HoverRoomDatabase) {
...
                 if (t == null) {
+                    isNew = true
                     c?.let { Utils.logAnalyticsEvent(c.getString(R.string.initializing_ussd_services), c) }
                     t = StaxTransaction(intent, a, contact, c)
                     transactionDao.insert(t)
                     t = transactionDao.getTransaction(t.uuid)
                 }
+
                 t!!.update(intent, a, contact, c)
+                if( !isNew && shouldFlagFailure(t, c)) t.setFailed_Incomplete()
```
davkutalek, line 169: Instead of doing all this work here, just add the logic for this directly to the update method of the staxtransaction object. That way the DatabaseRepo doesn't need to know about the internals of transactions.

davkutalek (changes requested):
Minor refactoring needed, otherwise I think its go to go

app/src/main/java/com/hover/stax/database/DatabaseRepo.kt
androidmaven: Thanks, changes made

app/src/main/java/com/hover/stax/database/DatabaseRepo.kt
androidmaven: Thanks, changes made

PR #332, opened 2021-09-16
178731662 multi account support

alexosugo:


davkutalek:
This looks great. The approach to multi balance checks looks solid.

The latestBalance, latestBalanceTimestamp, and accountNo fields should be removed from the Channel entity/DB. 

As noted by @androidmaven there are a few places where accounts still need to replace channels, but they look straightforward.

davkutalek:
From @androidmaven on Slack:

1. From AccountDetailsFragment -> Rename account -> Back to AcountDetailsFragment -> Click back from here crashed the app.
2. Rename account doesnt go past the string validation -> seems it has bug in its validation -> Always shows red highlight of “please enter a new account name”
3. Refresh balance from AccountDetailsFragment needs to use account instead of channel
4. From AccountDetailsFragment -> Select a transaction needs to use account instead of channel

PR #340, opened 2021-10-15
Fixes merge

alexosugo:


app/src/main/java/com/hover/stax/bounties/BountyViewModel.kt
```diff
@@ -83,25 +96,41 @@ class BountyViewModel(application: Application) : AndroidViewModel(application)
 
     private fun loadChannels(actions: List<HoverAction>?): LiveData<List<Channel>> {
         if (actions == null) return MutableLiveData()
-        val ids = getChannelIdArray(actions)
-        return repo.getChannels(ids)
+        val ids = getChannelIdArray(actions.distinctBy { it.id }).toList()
```
davkutalek, line 99: @alexosugo the reason this didn't help is that this is distinct by the action's id, which doesn't do anything. It should be distinct by `it.channel_id`, then you likely won't need to batch it.

PR #357, opened 2021-11-29
180467219 firebase with device

androidmaven:
- Extracted a new class called `AnayticsUtils `from the `utils.kt `to make it cleaner.
 - Cleaned up AnaytlicsUtils
 - Make FirebaseAnalytics set user Id as the device Id, so that its easy to target users from Firebase.

### DIFF
--- a/app/src/main/java/com/hover/stax/utils/AnalyticsUtil.kt
+++ b/app/src/main/java/com/hover/stax/utils/AnalyticsUtil.kt
@@ -1,4 +1,107 @@
 package com.hover.stax.utils
 
+import android.content.Context
+import android.os.Bundle
+import com.amplitude.api.Amplitude
+import com.amplitude.api.Identify
+import com.appsflyer.AppsFlyerLib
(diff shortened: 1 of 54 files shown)


app/src/main/java/com/hover/stax/utils/AnalyticsUtil.kt
```diff
@@ -0,0 +1,107 @@
...
+import com.google.firebase.ktx.Firebase
+import com.google.firebase.ktx.initialize
+import com.hover.sdk.api.Hover
+import com.hover.stax.BuildConfig
+import org.json.JSONException
+import org.json.JSONObject
+import timber.log.Timber
+import java.util.*
+
+class AnalyticsUtil {
```
alexosugo, line 19: You can change this to an `object` instead of a class. That way the underlying functions are static by default. You wouldn't need to wrap them in a `companion object`.

alexosugo (changes requested):

alexosugo (approved):

PR #371, opened 2021-12-14
bounty should always check for any successful transaction first, befo…

androidmaven:
…re setting for failed or pending state


app/src/main/java/com/hover/stax/bounties/Bounty.kt
```diff
@@ -12,7 +12,7 @@ class Bounty(val action: HoverAction, val transactions: List<StaxTransaction>) {
     val transactionCount get(): Int = transactions.size
 
     fun lastTransactionIndex(): Int = if (transactionCount == 0) 0 else transactionCount - 1
-
+    fun hasASuccessfulTransaction() : Boolean = transactions.find { it.status == Transaction.SUCCEEDED  } != null
```
alexosugo, line 15: You can simplify this to `transactions.any { it.status == Transaction.SUCCEEDED }`. The result is a boolean that satisfies that condition.

alexosugo (changes requested):

alexosugo (approved):

PR #376, opened 2021-12-20
fix for account not adding

androidmaven:
- I noticed the cause of the bug is that it tries to fetch if the action is already in a fetch_account list, of which not all account is has a type of fetch_account.

- I think the check is not necessary since the insert method already have a 
 ```
@Insert(onConflict = OnConflictStrategy.REPLACE)
    fun insert(account: Account)
```

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -8,6 +8,7 @@
   <component name="DesignSurface">
     <option name="filePathToZoomLevelMap">
       <map>
+        <entry key="../../../../layout/custom_preview.xml" value="0.24444444444444444" />
         <entry key="app/src/main/res/drawable/button_bg_colored.xml" value="0.32037037037037036" />
         <entry key="app/src/main/res/drawable/button_bg_dark.xml" value="0.32037037037037036" />
         <entry key="app/src/main/res/drawable/button_bg_grey.xml" value="0.32037037037037036" />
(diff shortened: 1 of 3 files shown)


davkutalek:
I don't think this is the issue. The issue is that the user tries to add channel(s) and whether or not they check balance/fetch accounts the channel(s) are not added to their list of selected channels. I believe the issue occurs when the configuration for the given channel is somehow invalid. It happens consistently for some channels and not others.

androidmaven:
@davkutalek ..... You're correct about the issue.

- And the problem from what I checked was in the if statement (getFetchAccount(t.id) == null), at this stage, if it returns null then the account won't save.
- Cos like you mentioned, the issue is at the creating of account stage, whether or not they check or fetch their account.

alexosugo:
@androidmaven That condition ensures that when adding new accounts, only the ones that don't support multiple accounts get saved e.g MMOs. The rest are added to the run list for fetching accounts. @davkutalek is right that the issue may be the configuration. Unless the fetch accounts action is run successfully and valid accounts parsed, they will not be created. 

Removing that condition means that an account will be created but without a valid account name. Then when running any action, we won't be able to tell which account it belongs to.

androidmaven:
@alexosugo @davkutalek , Agreed. So in this case, should we sort of force users to fetch accounts when adding an account for the first time? Or maybe show them a different dialogue content letting them "complete the process of adding their account?"

alexosugo:
The first action for any channel with multiple accounts is always `fetch_account`. I agree we need to inform the user. Maybe inform them that the `fetch account` action failed and ask them to retry.

PR #378, opened 2022-01-14
179527998 fill in extra variables

androidmaven:
- Show non-template variables
 - Support non-template variables on the summary card
 - Support validation of the variables and set states


app/src/main/java/com/hover/stax/home/MainActivity.kt
```diff
@@ -344,6 +345,12 @@ class MainActivity : AbstractNavigationActivity(), BalancesViewModel.RunBalanceL
                     .extra(HoverAction.NOTE_KEY, transferViewModel.note.value)
                     .extra(Constants.ACCOUNT_NAME, selectedAccount?.name)
 
+            if(nonTemplateVariables!=null && nonTemplateVariables.isNotEmpty()) {
```
alexosugo, line 348: You can use `!isNullOrEmpty()` to simplify the condition.

app/src/main/java/com/hover/stax/transfers/NonTemplateSummaryAdapter.kt
```diff
@@ -0,0 +1,40 @@
...
+
+    @SuppressLint("NotifyDataSetChanged")
+    fun updateList(key: String, value: String) {
+        itemMap.put(key, value)
+        notifyDataSetChanged()
+    }
+
+    inner class ViewHolder(val binding: SummarycardNontemplateItemsBinding) : RecyclerView.ViewHolder(binding.root) {
+        fun bindItems(key: String, value: String) {
+            binding.itemKey.text = key
```
alexosugo, line 20: Since the `itemKey` is not an inflatable view, it doesn't need to be part of the view/layout. If the `key` needs to be accessed, then the property in the `bindItems` function is enough. .

app/src/main/java/com/hover/stax/transfers/TransferViewModel.kt
```diff
@@ -116,6 +119,50 @@ class TransferViewModel(application: Application, repo: DatabaseRepo) : Abstract
         }
     }
 
+    fun initNonTemplateVariables(entries: List<String>) {
+        val itemList = ArrayList<NonTemplateVariable>()
```
alexosugo, line 123: Use a mapping function to initialise the `NonTemplateVariable` from the entries.

app/src/main/java/com/hover/stax/transfers/TransferViewModel.kt
```diff
@@ -116,6 +119,50 @@ class TransferViewModel(application: Application, repo: DatabaseRepo) : Abstract
...
+    }
+
+    fun nullifyNonTemplateVariables() {
+        nonTemplateVariables.postValue(null)
+    }
+    fun updateNonTemplateVariables(nonTemplateVariable: NonTemplateVariable) {
+        var itemList = nonTemplateVariables.value
+        if(itemList == null) itemList = ArrayList()
+
+        itemList.find { it.key == nonTemplateVariable.key }
```
alexosugo, line 137: You can combine the `find` and `let` with `firstOrNull`

alexosugo (changes requested):

app/src/main/java/com/hover/stax/transfers/NonTemplateSummaryAdapter.kt
androidmaven: Actually, the itemKey is an inflatable view, however it's badly named. I've changed the word to be itemLabel, not key.

app/src/main/java/com/hover/stax/home/MainActivity.kt
androidmaven: Done, thanks.

app/src/main/java/com/hover/stax/transfers/TransferViewModel.kt
androidmaven: Done, thanks. I like the map function. Cleaner.

app/src/main/java/com/hover/stax/transfers/NonTemplateSummaryAdapter.kt
androidmaven: Done

app/src/main/java/com/hover/stax/transfers/NonTemplateVariable.kt
```diff
@@ -0,0 +1,19 @@
+package com.hover.stax.transfers
+
+import java.util.*
+
+data class NonTemplateVariable(val key: String,
```
davkutalek, line 5: The fact this is empty is a code smell. Objects should have a purpose, and if they contain nothing that means either code that is their responsibility is being done elsewhere or they should not exist. In this case I think it is that it should not exist. 

The view state does not need to be stored: since we aren't storing the view state for the rest of the form this leads to different behavior from these fields than the other fields. View state can be recalculated if necessary. A LinkedHashMap can be used to store the variables in the view model which will actually simplify the logic in a number of places

app/src/main/java/com/hover/stax/transfers/NonTemplateVariableAdapter.kt
```diff
@@ -0,0 +1,82 @@
...
+                    editTextListener.nonTemplateVariableInputUpdated (nonTemplateVariable)
+                }
+            }
+
+            binding.variableInput.addTextChangedListener(inputTextWatcher)
+            binding.variableInput.setHint(nonTemplateVariable.key)
+            binding.variableInput.tag = nonTemplateVariable.key
+            binding.variableInput.text = nonTemplateVariable.value
+
+            nonTemplateVariable.editTextState?.let {
```
davkutalek, line 42: Instead of storing the state in an object just move the validation call to this class. Then it can set the state of each variable directly instead of reloading all the data after the validation is done.

app/src/main/java/com/hover/stax/transfers/TransferFragment.kt
```diff
@@ -227,7 +235,10 @@ class TransferFragment : AbstractFormFragment(), ActionSelect.HighlightListener
         val recipientError = transferViewModel.recipientErrors(actionSelectViewModel.activeAction.value)
         contactInput.setState(recipientError, if (recipientError == null) AbstractStatefulInput.SUCCESS else AbstractStatefulInput.ERROR)
 
-        return channelError == null && actionError == null && amountError == null && recipientError == null
+        val nonTemplateVariableHasError = transferViewModel.nonTemplateVariablesAnError()
```
davkutalek, line 238: As noted elsewhere, if you move this validation call from the viewmodel to the adapter it will simplify the logic some

davkutalek (changes requested):
Overall looks good. I made some comments regarding the how this could be done without the `NonTemplateVariable` object, but if you want to leave as is it is fine.

I would recommend the word "Standard" in place of "Template". Template is a word with a lot of meaning already and makes me want to go look for a template somewhere, whereas standard I think makes it clearer. "Nonstandard" is a widely understood term, "nontemplate" is a bit weird.

PR #385, opened 2022-01-21
180668782 mv1

androidmaven:
- MVT 1 Designs
- Refactored MainActivity to split several methods into respective abstract classes.


davkutalek (approved):
Looks good! I like the way you've broken Main Activity up, however it would be better if all the parts weren't Activities themselves. Remember to [prefer composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance). Having them all be activities leads to having to weirdly chain them. The alternative is to make them regular classes or interfaces and pass the activity in when you need it. Some of them might be fine as Activities, just not all of them.

I don't think we should have to store the google logo in our app?

PR #386, opened 2022-01-21
180668670 mv2

androidmaven:


davkutalek (approved):
Missing the actual variant choice logic, but otherwise LGTM

PR #391, opened 2022-01-27
Onboarding mvt1 3

androidmaven:
This is a consolidation of mvt1 and mvt2 into one branch, so that it can be merged easily to development.


alexosugo (approved):

app/src/main/java/com/hover/stax/ui/theme/Shape.kt
```diff
@@ -0,0 +1,11 @@
+package com.hover.stax.ui.theme
+
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material.Shapes
+import androidx.compose.ui.unit.dp
+
+val Shapes = Shapes(
```
androidmaven, line 7: Maybe call this RoundedCornerShapes?

alexosugo: This can be extended to include other shapes, not just the ones for rounded corners.

PR #392, opened 2022-01-27
make wellness tips and banner's medium page accessible on settings page

androidmaven:


app/src/main/java/com/hover/stax/settings/SettingsFragment.kt
```diff
@@ -113,6 +114,13 @@ class SettingsFragment : Fragment(), NavigationInterface {
         }
     }
 
+    private fun setupLearnCard() {
+        with(binding.staxLearn) {
+            learnFinances.setOnClickListener {  findNavController().navigate(R.id.action_navigation_home_to_wellnessFragment) }
```
alexosugo, line 119: The navigation action here is incorrect. The source fragment here is the `settings`.

app/src/main/java/com/hover/stax/settings/SettingsFragment.kt
```diff
@@ -113,6 +114,13 @@ class SettingsFragment : Fragment(), NavigationInterface {
         }
     }
 
+    private fun setupLearnCard() {
```
alexosugo, line 117: The learn card isn't showing up in settings for me.

alexosugo (changes requested):
Navigation to the USSD Lib fragment also appears to be broken.

app/src/main/java/com/hover/stax/settings/SettingsFragment.kt
androidmaven: New push should fix

androidmaven:
Thanks for the correction, @alexosugo

PR #394, opened 2022-02-01
Add support for bounties that are cross border

davkutalek:
Fix for using some of the SDK's android variables. Requires https://github.com/UseHover/hover-sdk/pull/64


androidmaven (approved):
Looks good as well. Thanks

alexosugo (approved):

PR #399, opened 2022-02-09
Channels Update

alexosugo:
- removes all channels before updating the list on the workers
- changes the update channels worker to a coroutineworker


androidmaven (approved):
Looks good.

PR #402, opened 2022-02-15
logout if login fail

alexosugo:
- Logs out the user if login process fails at any point.
- Removes error message shown before referral details are populated on referral dialog. 
- Sends GoogleSignInAccount `idToken` to server instead of the firebase one to backend.


androidmaven (commented):
Due to styling diffs, I'm unable to find the exact change of using GoogleSignIn idToken from Firebase.

Could you please point me to it?

app/src/main/java/com/hover/stax/login/LoginViewModel.kt
```diff
@@ -30,189 +33,192 @@ private const val BOUNTY_EMAIL_KEY = "email_for_bounties"
...
+            onError(application.getString(R.string.login_google_err))
+        }
+    }
+
+    private fun signIntoFirebase(idToken: String, activity: AppCompatActivity) {
+        val credential = GoogleAuthProvider.getCredential(idToken, null)
+        auth.signInWithCredential(credential).addOnCompleteListener(activity) {
+            progress.value = 33
+            if (it.isSuccessful) {
+                auth.currentUser?.let { user -> setUser(user, idToken) }
```
alexosugo, line 74: I added the `idToken` here as a parm.

app/src/main/java/com/hover/stax/login/LoginViewModel.kt
```diff
@@ -30,189 +33,192 @@ private const val BOUNTY_EMAIL_KEY = "email_for_bounties"
...
+        progress.postValue(100)
+        saveResponseData(json)
+    }
+
+    private fun setUser(firebaseUser: FirebaseUser, idToken: String) {
+        Timber.e("setting user: %s", firebaseUser.email)
+        user.postValue(firebaseUser)
+        setEmail(firebaseUser.email)
+
+        uploadUserToStax(firebaseUser.email, idToken)
```
alexosugo, line 156: I removed the firebase `token` fetch here and called the `uploadUserToStax` method directly.

app/src/main/java/com/hover/stax/login/LoginViewModel.kt
androidmaven: Got it.

app/src/main/java/com/hover/stax/login/LoginViewModel.kt
androidmaven: Got it, thanks!

We probably might want to refractor the login view model a little bit in another story.
For example: SignInFirebaseAsync is doing like >3 things other than just signing in. 

Just an observation while reading.

This particular work is great, thanks.

androidmaven (approved):

PR #404, opened 2022-02-17
hides paybill for all other countries other than Kenya

alexosugo:


androidmaven (approved):
This is good to go for now.

We might make a separate story to make pay bill supported country-codes feed on Firebase instead of hardcoded.

PR #405, opened 2022-02-17
181283888 fix to non standard variables

androidmaven:
One thing is not clear.

RecipientInstituiton coming in from the API as a variable means it needs to be a filled value for the SDK to run whether it's visible to the user or not.

The code changes in this attached branch, making it non-visible for users, however, it doesn't run without setting "RecipientInstitution" a value. 

What should be auto-filled? the RecipientInstitution ID or name? I've included the recipient institution ID for this PR. Kindly let me know your thoughts in the review comments.

FYI:
The bug "Fix crash preventing the transaction from starting" was fixed earlier on in some other branch merged already.


alexosugo (changes requested):
I still get a crash after filling in the form and hitting continue.

androidmaven:
@alexosugo , fixed the bug

app/src/main/java/com/hover/stax/actions/ActionSelectViewModel.kt
```diff
@@ -32,48 +35,22 @@ import com.hover.stax.R
...
-        val map: LinkedHashMap<String, String> = LinkedHashMap()
-        variableKeys.forEach {
-            map[it] = ""
+            if (!isStandardVariable(it)) variableMap[it] = ""
         }
-        nonStandardVariables.postValue(map)
+        nonStandardVariables.postValue(variableMap)
     }
 
-    private fun nullifyNonStandardVariables() {
```
davkutalek, line 70: You don't need this since the function above will return an empty list

app/src/main/java/com/hover/stax/contacts/PhoneHelper.java
```diff
@@ -12,19 +12,6 @@
 public class PhoneHelper {
     final static private String TAG = "PhoneHelper";
 
-    public static String getNumberFormatForInput(String accountNumber, HoverAction a, Channel c) {
-        try {
-            String format = a.getFormatInfo(HoverAction.PHONE_KEY);
```
davkutalek, line 17: This is usually empty

app/src/main/java/com/hover/stax/contacts/PhoneHelper.java
```diff
@@ -12,19 +12,6 @@
 public class PhoneHelper {
     final static private String TAG = "PhoneHelper";
 
-    public static String getNumberFormatForInput(String accountNumber, HoverAction a, Channel c) {
```
davkutalek, line 15: Don't pass complex objects to a helper function. Now it has to know about the internals of 2 objects

app/src/main/java/com/hover/stax/home/AbstractSDKCaller.kt
```diff
@@ -121,15 +120,9 @@ abstract class AbstractSDKCaller : AbstractGoogleAuthActivity(), PushNotificatio
 
     private fun addRecipientInfo(hsb: HoverSession.Builder) {
         hsb.extra(HoverAction.ACCOUNT_KEY, transferViewModel.contact.value!!.accountNumber)
-                .extra(
-                        HoverAction.PHONE_KEY, PhoneHelper.getNumberFormatForInput(
```
davkutalek, line 125: Don't do normalization here, this is what the HoverSession class is for which already doing normalization on the phone number

davkutalek (approved):
Refactored some, please review

davkutalek:
@alexosugo updated the above but can't test due to build issues

PR #428, opened 2022-03-20
v.1.9.0.8

alexosugo:
- New fonts (Latino-type Brutalista).
- Updates logo on home page. 
- Uses account placeholders when accounts not fetched during setup. 
- Logout option in settings. Shows who the user is logged in as. 
- Username updated during Google login. 
- Fixes padding on Google sign in button on bounties. 
- Fixes invisible cursor on text inputs. 
- Basic permissions dialog design updated.


davkutalek:
@alexosugo so glad you were able to make it so we don't have to set font-normal everywhere. What was causing that?

alexosugo:
That was because of how the fonts were being loaded in Android. Before, we were using the `Lingver` library with a custom font declaration. Now, the fonts are being loaded from a fonts folder. The base font for all text views is set in the theme and is applied app-wide. We now only need to override it in places we need to set a bold or medium font. Only thing we change to achieve this is to use the `app` namespace instead of `android` when setting `fontFamily` on a textview. Preview will show the components with the set font and you can edit as needed. However, you need to switch the theme on the layout preview from `Starting` to `StaxTheme` to see the changes.

PR #430, opened 2022-03-23
181524654 show balance by default

androidmaven:


app/src/main/java/com/hover/stax/utils/Utils.kt
```diff
@@ -45,6 +45,11 @@ object Utils {
         return getSharedPrefs(c).getBoolean(key, false)
     }
 
+    fun getBooleanDefaultTrue(key: String?, c: Context): Boolean {
```
alexosugo, line 48: Since this method is only used once, you can add a default argument to the `getBoolean()` method to reduce duplication.

alexosugo (changes requested):

app/src/main/java/com/hover/stax/utils/Utils.kt
androidmaven: Done, thanks.

alexosugo (approved):

PR #431, opened 2022-03-25
181534590 update default account

alexosugo:
Default account updated successfully from settings.


androidmaven (approved):
Looks good to me.

A question though, why replace LiveData with Flow when they both do the same thing?

alexosugo:
@androidmaven Flows, unlike LiveData allow us to receive automatic updates asynchronously whenever there are any database changes. We don't need to re-request for the data during an activity/fragment's lifecycle.

PR #433, opened 2022-03-29
181712053 nav fix

alexosugo:
- adds safeargs plugin
- navigation for problematic fragments is through fragment directions. 
- added helper function to always validate the current destination before navigating.

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -9,6 +9,7 @@
     <option name="filePathToZoomLevelMap">
       <map>
         <entry key="../../../../layout/compose-model-1647504975633.xml" value="0.2673141891891892" />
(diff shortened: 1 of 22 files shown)


davkutalek (commented):
Google is really going library crazy. This is much better with safe args tho.

The only issue is that this adds a 3rd class to our existing navigation helper classes when we really should have just one. Can we consolidate them?

alexosugo:
@davkutalek Yes. We need to simplify navigation a lot further. I intend to remove the 2 extra helper classes `NavigationInterface` and `StaxNavigation`. All navigation will be encapsulated in the respective fragments.

androidmaven (approved):
LGTM. Thanks for this.

PR #434, opened 2022-03-29
Bugfix/1.9.1

alexosugo:


app/src/main/java/com/hover/stax/transfers/AbstractFormFragment.kt
```diff
@@ -158,11 +158,11 @@ abstract class AbstractFormFragment : Fragment(), AccountDropdown.AccountFetchLi
...
 
     private fun runBalanceCheck(channelId: Int) = lifecycleScope.launch(Dispatchers.IO) {
-        channelsViewModel.getChannel(channelId)?.let {
+        channelsViewModel.getChannel(channelId)?.let { channel ->
             val action = channelsViewModel.getFetchAccountAction(channelId)
-            channelsViewModel.setActiveChannel(it)
+            channelsViewModel.setActiveChannel(channel)
 
-            (activity as? MainActivity)?.makeCall(action!!, it)
+            action?.let { a -> (activity as? MainActivity)?.makeCall(a, channel) }
```
davkutalek, line 165: @alexosugo did this throw a nullpointer? Seems very bad if it did, action shouldn't be null right?

davkutalek: Should at least show an error if this happens right?

alexosugo: Yes, it did. I've encountered two instances where action should not be null but an npe is thrown. Here and on the TransactionReceiver.

PR #435, opened 2022-03-29
180668675 search accounts

androidmaven:
- Ability to search for accounts.
- Shows empty states when no account is found for a keyword.


alexosugo (changes requested):
Code changes look good. However, a few things:
- When no sim card is installed, it shows the empty view when an invalid search params is entered. 
- No channels are shown when no sim is present. 
- The text on the button should not be two lines. 
- The CTA performs no action.

alexosugo (approved):

PR #438, opened 2022-03-31
181725890 nav updates

alexosugo:
This was a refactor of the app's navigation. The three problems it sought to address are:
- Navigation was split into multiple classes, some with redundant functions.
- Most of the routing passes through the MainActivity and HomeFragment. 
- Invalid routes when navigating between fragments. 

We now only have two classes:
- NavUtil:  contains helper functions to validate origin and destination fragments before navigating. Also has a function to inflate TransactionDetails from anywhere in the app.
- NavHelper (renamed from StaxNavigation): initializes `MainActivity` navigation, checks permissions and handles all nav functions declared in `MainActivity`.

Changes here include:
- Removing the `NavigationInterface`.
- Stripping the `StaxNavigation` of all nav functions that don't require permissions and keep only the ones that originate from the `MainActivity` class. 
- All fragments handle their own navigation.


androidmaven (changes requested):
Code updates looks sleek.

Some functional updates tho.

None of the DeepLinking's Navigation is firing.

alexosugo:
@androidmaven @davkutalek These commits introduce changes to how the app handles deep links. Deep links are now declared at fragment level in the navigation graph like below:

    <fragment
            android:id="@+id/navigation_transfer"
            android:name="com.hover.stax.transfers.TransferFragment"
            android:label="@string/nav_transfer"
            tools:layout="@layout/fragment_transfer">
        <action
                android:id="@+id/action_navigation_transfer_to_accountsFragment"
                app:destination="@id/accountsFragment" />
        <action
                android:id="@+id/action_navigation_transfer_to_navigation_home"
                app:destination="@id/navigation_home"
                app:popUpTo="@id/navigation_home"
                app:popUpToInclusive="true" />
        <argument
                android:name="transactionType"
                app:argType="string" />
        <deepLink
                android:id="@+id/transferDL"
                app:mimeType="text/plain"
                app:uri="stax.me/transfer/{transactionType}" />
    </fragment>

Android now handles the deep links automatically whenever a valid URI is clicked outside of Stax. 

In the example above, the deep link handles routing to the `TransferFragment` allowing us to specify what type of transfer action we want to execute. Since we're using `safeargs`, the `transactionType` has to be passed so that the fragment can load successfully. We cannot use the previous method of using two deep links and since Android handles the navigation for us. Therefore, all `send money` and `buy airtime` deep links have to be updated to reflect this. 

An advantage of this, is that we can declare routes quite easily like we do on web e.g `https://stax.me/{channels}/{country}/{type}` if we want.

app/src/main/java/com/hover/stax/utils/Utils.kt
```diff
@@ -264,4 +265,18 @@ object Utils {
         else
             UIHelper.flashMessage(c, c.getString(R.string.enable_call_permission))
     }
+
+    fun openSupportEmailClient(activity: Activity) {
```
davkutalek, line 269: This is a near duplicate of openEmail above. Do we need both?

davkutalek (approved):
This is great, the deep links are so much better this way

androidmaven (approved):
Fantastic work

app/src/main/java/com/hover/stax/utils/Utils.kt
alexosugo: Yeah. It's a duplicate. I've removed one.

PR #440, opened 2022-04-04
181584842 search ussd lib

androidmaven:
- New feature: searching of ussd library channels.
- Made some minor refractor to the search of ChannelsViewModel to prevent code duplication


app/src/main/java/com/hover/stax/channels/Channel.java
```diff
@@ -97,6 +97,16 @@ public Channel(JSONObject jsonObject, String rootUrl) {
         return sorted_list;
     }
 
+    public static List<Channel> filter(List<Channel> channels, String searchValue) {
```
alexosugo, line 100: Preferably let's keep models clean. Logic like this can be moved to a ViewModel or where it's being used.

alexosugo (changes requested):
Search works well on the USSD library, however, make the search adhere to the search param strictly. Search by country name doesn't change the result. The country is already handled by the country filter. If no sim is installed, then search by country can be used.

app/src/main/java/com/hover/stax/channels/Channel.java
davkutalek: Agreed. Even the toFilterableString method below doesn't belong here. I think the root of the problem is the ChannelViewModel doing too much. Possibly need a specialized channel search view model to reduce duplication properly

davkutalek (approved):
LGTM, once kombo is happy I think it is good to go. We can handle refactoring of channel view model separately

davkutalek:
> Search works well on the USSD library, however, make the search adhere to the search param strictly. Search by country name doesn't change the result. The country is already handled by the country filter. If no sim is installed, then search by country can be used.

New designs have a country dropdown, so should be totally separate. We'll do a new story to implement, for now just keep it simple and leave country out of the search

app/src/main/java/com/hover/stax/channels/Channel.java
androidmaven: Done

app/src/main/java/com/hover/stax/views/ChannelEmptyStateView.kt
```diff
@@ -0,0 +1,37 @@
...
+import android.content.Context
+import android.text.method.LinkMovementMethod
+import android.util.AttributeSet
+import android.view.LayoutInflater
+import android.widget.LinearLayout
+import androidx.core.text.HtmlCompat
+import com.hover.stax.R
+import com.hover.stax.databinding.NoAccountEmptyStateBinding
+
+class ChannelEmptyStateView (context: Context, attrs: AttributeSet) : LinearLayout(context, attrs) {
```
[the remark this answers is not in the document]
androidmaven: Done

davkutalek (approved):
This looks good to me. There are some conflicts with development branch that need to be worked out

PR #445, opened 2022-04-10
Info system component

androidmaven:


app/src/main/java/com/hover/stax/utils/Utils.kt
```diff
@@ -246,6 +245,17 @@ object Utils {
         }
     }
 
+    fun getActivity(context: Context): Activity? {
```
alexosugo, line 248: If you find yourself needing to do this, then you've complicated your solution. Context should be injected, inferred or passed in most cases.

app/src/main/java/com/hover/stax/views/StaxInfoTextView.kt
```diff
@@ -0,0 +1,46 @@
...
+
+import android.content.Context
+import android.util.AttributeSet
+import android.view.LayoutInflater
+import android.widget.FrameLayout
+import com.hover.stax.R
+import com.hover.stax.databinding.InfoTextLayoutBinding
+import com.hover.stax.utils.Utils
+
+class StaxInfoTextView(context: Context, attrs: AttributeSet) : FrameLayout(context, attrs) {
```
alexosugo, line 11: Use plain XML to simplify this. It's a simple component that doesn't introduce any new complex functionality. Consider XML layouts first before creating custom components.

app/src/main/java/com/hover/stax/views/StaxInfoTextView.kt
```diff
@@ -0,0 +1,46 @@
...
+			textValue = a.getString(R.styleable.StaxInfoTextView_android_text) ?: ""
+		} finally {
+			a.recycle()
+		}
+	}
+
+	private fun fillAttr() {
+		binding.textId.text = textValue
+		binding.root.setOnClickListener {
+			Utils.getActivity(context)?.let {
```
alexosugo, line 36: Activity/fragment lifecycle may impact access to context here. See comment above on context in Utils.

alexosugo (changes requested):

davkutalek (approved):

PR #446, opened 2022-04-11
181700347 marketing lists

alexosugo:
This PR introduces a `StaxUser` object which is persisted in the database whenever a user logs in or updates their account. The `is_mapper` flag can now be checked reliably after login to be able to show the user the appropriate state and message when they go to bounties. 

Also required is the `marketing_opt_in` flag to show the correct checkbox state in Settings. We can save this in preferences but it will not survive between installs.


davkutalek (approved):
Refactor is great!

davkutalek:
@alexosugo added `marketing_opted_in` to server response, now live on development

alexosugo:
Changes added factor in marketing opt in flag in the stax user object. 
- When settings fragment is loaded, the marketing checkbox state is set by the flag. 
- When a user is not signed in and they opt in to marketing, they are prompted to sign in with Google. 
- When a user signs out, the marketing checkbox is unchecked.

PR #448, opened 2022-04-12
Click Retry transaction should autoFill transfer fragment

androidmaven:


app/src/main/java/com/hover/stax/transfers/TransferFragment.kt
```diff
@@ -192,6 +204,12 @@ class TransferFragment : AbstractFormFragment(), ActionSelect.HighlightListener,
             }
         }
     }
+    private fun observeContact() {
+        transferViewModel.contact.observe(viewLifecycleOwner){
+            Timber.i("contact is ${it?.shortName()} at observing level")
```
androidmaven, line 209: @davkutalek something is weird at this point. Need some help why.

How to reproduce.
Click on Try again the first time - logger shows expected value.
Click back and click on try again subsequently - contact shows null, even though at the point where it updates the contact live data is not null.

This seems to also be the bug causing.
-contact name showing as null in the transaction history after loading it subsequently.

app/src/main/java/com/hover/stax/requests/Request.kt
```diff
@@ -0,0 +1,225 @@
...
+	companion object {
+		const val PAYMENT_LINK_SEPERATOR = "-"
+		private const val TAG = "Request"
+		fun isShortLink(link: String): Boolean {
+			return link.length <= 25
+		}
+
+		val encryptionSettings: Encryption.Builder
+			get() = Encryption.Builder().setKeyLength(128).setKeyAlgorithm("AES")
+				.setCharsetName("UTF8").setIterationCount(65536).setKey("ves€Z€xs€aBKgh")
```
johnswekenyi, line 167: Should we be passing this here? could we make it a variable? @davkutalek thoughts?

johnswekenyi (commented):
LGTM

app/src/main/java/com/hover/stax/requests/Request.kt
davkutalek: @johnswekenyi yeah, there are serious issues with this current approach. We did it this way to work offline and have been planning to re-do it but waiting since planning full redesign. Lumi has been working on the new designs for us, so we should get to it soon-ish.

davkutalek (commented):
This looks mostly ok to me. I have some small problems with the way the transfer fragment is loading data, tho they were mostly already present from requests. I also think there may be a slightly better way to do the navigation, but will let @alexosugo weigh in on that.

davkutalek (approved):
LGTM

app/src/main/java/com/hover/stax/futureTransactions/RequestsAdapter.kt
```diff
@@ -0,0 +1,59 @@
...
+		holder.itemView.setOnClickListener { selectListener.viewRequestDetail(r.id) }
+	}
+
+	private fun shouldShowDate(r: Request, position: Int): Boolean {
+		return position == 0 || humanFriendlyDate(requestList!![position - 1].date_sent) != humanFriendlyDate(
+			r.date_sent)
+	}
+
+	override fun getItemCount(): Int {
+		return if (requestList != null) requestList!!.size else 0
```
alexosugo, line 42: This can be simplified with an elvis operator.

app/src/main/java/com/hover/stax/futureTransactions/RequestsAdapter.kt
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	override fun getItemId(position: Int): Long {
+		return position.toLong()
+	}
+
+	override fun getItemViewType(position: Int): Int {
```
alexosugo, line 49: We don't usually need this in a recyclerview adapter unless we're inflating different view types.

app/src/main/java/com/hover/stax/futureTransactions/RequestsAdapter.kt
```diff
@@ -0,0 +1,59 @@
...
+	private fun shouldShowDate(r: Request, position: Int): Boolean {
+		return position == 0 || humanFriendlyDate(requestList!![position - 1].date_sent) != humanFriendlyDate(
+			r.date_sent)
+	}
+
+	override fun getItemCount(): Int {
+		return if (requestList != null) requestList!!.size else 0
+	}
+
+	override fun getItemId(position: Int): Long {
```
alexosugo, line 45: This isn't required unless we need to use the ID elsewhere or if we're using a selectable recyclerview.

app/src/main/java/com/hover/stax/requests/Request.kt
alexosugo: I'm generally not a big fan of having functions inside model classes other than constructors just to keep them clean. Functionality can be moved out to a view model or helper class. Thoughts @johnswekenyi @davkutalek

alexosugo (commented):
Clicking retry on a failed airtime txn loaded the form as intended, however, the wrong account was used. The originating account was my bank but my default account was picked; MPESA.

app/src/main/java/com/hover/stax/futureTransactions/RequestsAdapter.kt
androidmaven: I've encountered scenarios where the visible items show incorrect data after scrolling up and down. A stackOverFlow answer suggested this answer and worked.

I'm unable to find the link cos it's a long long find. Maybe recent Android updates come with this fix built-in.

app/src/main/java/com/hover/stax/futureTransactions/RequestsAdapter.kt
androidmaven: Kindly see previous comment. Thakns.

app/src/main/java/com/hover/stax/futureTransactions/RequestsAdapter.kt
androidmaven: Done, thanks!

app/src/main/java/com/hover/stax/requests/Request.kt
davkutalek: @alexosugo the encryption could be moved to a helper class so that the request object doesn't have to worry about how the encryption is done, just calls methods to encrypt/decrypt. It is fundamental to the request object, so it should not be in a view model since those should deal with how the app works functionally from the user's point of view.

alexosugo: @davkutalek Agreed. However, the generate methods can be moved out. They don't impact the object in any way.

PR #450, opened 2022-04-14
complete task for bug: Pressing cancel button on PIN stage UX

androidmaven:
Story info:
Pressing back button on while on the PIN screen clears input and user has to go back and re-input all the information again


app/src/main/java/com/hover/stax/home/MainActivity.kt
```diff
@@ -226,6 +227,15 @@ class MainActivity : AbstractRequestActivity(), BalancesViewModel.RunBalanceList
         Timber.e("received result. %s", data?.action)
         Timber.e("uuid? %s", data?.extras?.getString("uuid"))
 
+        if(resultCode == RESULT_OK) handlePostSDK(requestCode, resultCode, data)
```
davkutalek, line 230: this needs to be properly integrated into the when statement below. It should really be in `showPopUpTransactionDetailsIfRequired` and should apply to any transaction, not just transfers.

davkutalek (changes requested):

app/src/main/java/com/hover/stax/home/MainActivity.kt
androidmaven: Agreed and effecting this: "properly integrated into the when statement"

I'm not quite sure @davkutalek, if "should apply to any transaction, not just transfers." because:

- Balance most likely would have been initiated from the home screen, not the transfer fragment.

app/src/main/java/com/hover/stax/home/MainActivity.kt
```diff
@@ -227,6 +228,10 @@ class MainActivity : AbstractRequestActivity(), BalancesViewModel.RunBalanceList
         Timber.e("uuid? %s", data?.extras?.getString("uuid"))
 
         when {
+            resultCode == RESULT_CANCELED && requestCode == Constants.TRANSFER_REQUEST -> {
```
davkutalek, line 231: This still isn't correct. It would override the scheduling case and not apply to balances. As stated before, it should be integrated into the showPopUpTransactionDetailsIfRequired method.

davkutalek (changes requested):

app/src/main/java/com/hover/stax/home/MainActivity.kt
androidmaven: Done, thanks!

davkutalek (approved):
LGTM

PR #452, opened 2022-04-21
181642714 cursor on account number

alexosugo:
Fix was quite simple actually; the dropdown had been set as `editable = false` by default.


androidmaven (approved):
lgtm

PR #457, opened 2022-04-25
181937962 share excitement

androidmaven:
- Match pivotal story


app/src/main/java/com/hover/stax/transactions/TransactionDetailsFragment.kt
```diff
@@ -49,16 +50,13 @@ class TransactionDetailsFragment : DialogFragment(), Target{
     private var uuid: String? = null
     private var isFullScreen = false
 
+    private lateinit var bottomSheetBehavior: BottomSheetBehavior<RelativeLayout>
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        isFullScreen = requireArguments().getBoolean(IS_FULLSCREEN)
-        if (isFullScreen) setFullScreen()
+        setFullScreen()
```
alexosugo, line 57: If transaction details will always be shown as full screen, let's convert this to a proper fragment. It can be simplified to remove unnecessary code.

alexosugo (changes requested):
The bottomsheet with the gif doesn't display correctly. See below video. 

https://user-images.githubusercontent.com/59045329/165497406-727914bd-cae5-4906-be02-86a542267bfa.mp4

davkutalek (changes requested):
This looks fine codewise. You need to get approval from marketing/graphics for something like this - its a pretty big statement. Also I don't really think its worth the size, looks like it would add ~350KB with just the one gif. But ask marketing what they want to show.

androidmaven:
Thanks, reverting the codebase to match actual spec. Will request for design thoughts on GIFs as a follow up item.

alexosugo (approved):

PR #460, opened 2022-04-28
complete refractoring sdk related activity results

androidmaven:
Kindly watch this 2mins video: https://www.loom.com/share/1b2ae90fda084fd689caadf9e813938d

This branch also fixes: `pressing cancel on the pin stage` to function by regular back stack.

>>SDKBuilder was later renamed to SDKIntent


davkutalek (changes requested):
I think conceptually this is a better approach to start Hover directly from the fragments. However the implementation here needs work since there are now 3 classes to deal with calling Hover (HoverSession, SDKIntent, SDKLaunchers). They know about the internals of all kinds of things that they shouldn't. There is also significant duplicated code between each Fragment that calls Hover. That kind of duplication leads to hard to notice bugs. 

See the refactor I did https://github.com/UseHover/Stax/pull/461 and especially [AbstractHoverCallerActivity](https://github.com/UseHover/Stax/pull/461/files#diff-49a8850d571002cc655c1329c5ae22bfcffe16d97bb2e732366412c41bc1d46a). That class could become an abstract fragment instead (of course moving some of the code such as the activity result listeners to the subclasses). `HoverSession.Builder` can become `HoverIntent.Builder`

A couple things to note: 1) Fragments and Activities get different requestCodes as results depending on which makes the call (meaning: if a fragment makes the call, its activity can catch the result but it will get a totally random different request code). That may be fine. 2) There *may* be context issues when doing it this way - if so the result from Hover may come immediately and not include any data. This may be fine as well.

davkutalek:
Just realized you talk about the issue with the Activity result in your video. The result not firing is the reason that we do this in the activity - Android doesn't handle the results between fragment calls and Activity calls properly. Because of that I think this PR should be closed. The solution in https://github.com/UseHover/Stax/pull/461 handles the de-duplication better.

PR #461, opened 2022-04-29
Channel and Repo breakdown

davkutalek:
This refactor started as cleaning up how requests are made to the Hover SDK as well as breaking up the ChannelViewModel and the DatabaseRepo. This ended up affecting almost every major part of the app especially because I realized that the focus of much of the app on Channels is unnecessary: In most cases we only care about the accounts the user has, not the channels they have selected. Because of how fundamental this change is, it seemed like the right time to fix other issues that I found along the way, leading to refactors of many of the fragments as well.

I have likely introduced a lot of new bugs, so this will need thorough testing. It also could likely use optimization - clean code was my goal, not the absolute most efficient code, so likely room for improvement. Fortunately this should be easier now since things are smaller. There are a number of fields that need to be removed from the DB, we may need to create a manual migration or set up a db asset and set fallback to destructive.

The paybill fragment in particular took quite a lot of refactoring due to the weird issues with our custom input layouts and back navigation. I was finally able to find the issue and fix it (see comment). Once that was fixed there was a lot to refactor in the paybill fragment since it was working around those issues. This is def the most complicated screen/flow in the app, so it needs particular attention.

Requires pulling and building the latest SDK code on the stax branch.


app/src/main/java/com/hover/stax/inapp_banner/BannerUtils.kt
```diff
@@ -14,7 +15,7 @@ import kotlin.properties.Delegates
 
 class BannerUtils(val context: Context) : KoinComponent {
 
-    private val repo: DatabaseRepo by inject()
+    private val repo: AccountRepo by inject()
```
davkutalek, line 18: Is this safe?

davkutalek:
Found the issue with our input components which is described [here](https://stackoverflow.com/questions/45818015/restore-fragment-with-two-views-having-the-same-id) and [here](http://web.archive.org/web/20180625034135/http://trickyandroid.com/saving-android-view-state-correctly/). Working on fixing and cleaning it up.

app/src/main/java/com/hover/stax/inapp_banner/BannerUtils.kt
alexosugo: Yes, it is; as long as it's been declared in the module.

davkutalek:
@alexosugo Managed to finish up pretty much everything with a few issues left:
- Add account in accounts dropdown works correctly, but it should have a plus icon and a divider line above it
- If the default account has never checked a balance it asks to check it when the user visits the home screen. This is fine, but sometimes it asks when it shouldn't, such as when a transaction is about to run. I actually think we should not be returning to the home screen once an SDK request is submitted, there should be some sort of "background" fragment that shows a loading message.
- The balance actions don't always seem to be loading correctly. This happens on the home screen and also when a user tries to transact with an account that is a placeholder (it correctly shows a message that they need to connect the account first, it just doesn't run the action).

davkutalek:
@alexosugo merged to latest development. However, I didn't have time to fully sort/check the conflicts with the list updates you made. Recommend re-basing that branch on this one. Should be easy, the library adapter was deleted cause it is identical to channel adapter and the bounty adapter should be fine, wasn't touched by this refactor.

PR #463, opened 2022-05-04
182024752 permissions fix

alexosugo:
This PR should fix the app flickering issue when accessibility permissions are granted in device settings. The issue was brought about by a condition that checks if `all` permissions have been granted before showing the accessibility dialog. This assumed that the `PHONE` and `SMS` permissions had already been granted at this stage. Toggling `PHONE` or `SMS` off from settings causes it to go into an endless loop.

The major change introduced here is how permissions are handled in the app settings (Stax). Here's some context from the developer docs:

_Starting in Android 11 (API level 30), if the user taps Deny for a specific permission more than once during your app's lifetime of installation on a device, the user doesn't see the system permissions dialog if your app requests that permission again. The user's action implies "don't ask again." On previous versions, users would see the system permissions dialog each time your app requested a permission, unless the user had previously selected a "don't ask again" checkbox or option. If a user denies a permission request more than once, this is considered a permanant denial._ 

Phone and SMS toggles in settings will redirect the user to device settings if the permission check comes back as either `GRANTED` or `DENIED`. Showing of the rationale is handled as well. The fallback state is requesting permissions as usual. This will work for all devices. 

We could check by device API level but that introduces complexity if we have to manage 
 - devices running Android 11 and above
 - devices below Android 11 that have selected `Dont Ask Again` in the permission dialog.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -66,9 +66,6 @@
                 android:launchMode="singleTask"
                 android:theme="@style/StaxTheme" />
         <activity
-                android:name=".destruct.SelfDestructActivity"
-                android:exported="false" />
-        <activity
                 android:name=".onboarding.OnBoardingActivity"
                 android:exported="false"
                 android:theme="@style/StaxTheme" />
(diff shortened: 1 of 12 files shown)


androidmaven (approved):
QA'ed, and bug fixed.

PR #468, opened 2022-05-05
Glide Intergration

alexosugo:
- replaces Picasso with Glide to improve image loading, performance and curb leaks
- removes cursor from account dropdowns

Note: It seems the SDK is dependent on Picasso. We need to change it to use Glide then remove the dependency.


androidmaven (approved):
I support the use of Glide, more because it allows us to load GIFs.

Looks good.

PR #474, opened 2022-05-11
uxcam integration and screen naming

mijiga:


app/src/main/java/com/hover/stax/permissions/PermissionsFragment.kt
```diff
@@ -40,6 +40,11 @@ class PermissionsFragment : DialogFragment() {
         return dialog!!.createIt()
     }
 
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
```
androidmaven, line 43: Seems its empty, maybe to remove?

androidmaven (changes requested):
Code additions look excellent and straightforward, but just a minor however vital request.

Could we make the UXCam log consistently with other existing logs?

For example in all screens (fragments) we already have the `AnalyticsUtil.logAnalyticsEvent(string, context)` which logs for firebase, amplitude and appsflyer.

I think inside the functions would be the perfect place to add the UXCam.tagScreenName(); without the need for extra new strings.

However, this `AnalyticsUtil.logAnalyticsEvent(string, context)` is also used for other logs other than "visited screen" so we might want add an extra flag in the function for example `AnalyticsUtil.logAnalyticsEvent(string, context, isForScreenVisit: Boolean? = false)`; and pass "true" inside each "visit logs"; this is so that we can appropriately check if we should log or not the UXCam in the `AnalyticsUtil.logAnalyticsEvent`.

Please let me know if you need further clarifications or its clear.
Thanks

mijiga:
@androidmaven makes sense, except when I do that the function calls without the `isForScreenVisit` variable set end up calling the`logAnalyticsEvent(event: String, context: Context, excludeAmplitude: Boolean)` func. So maybe should just create another `logAnalyticsEvent` func with different variable order?

androidmaven:
> @androidmaven makes sense, except when I do that the function calls without the `isForScreenVisit` variable set end up calling the`logAnalyticsEvent(event: String, context: Context, excludeAmplitude: Boolean)` func. So maybe should just create another `logAnalyticsEvent` func with different variable order?

I see your point @mijiga . Instead of going through that change. You could do something like

logAnalyticsEvent(event: String, ...:Object){
val firstWord = event.split(" ").first()
		if(firstWord == "visited") {
			//then log UXCam
		}
}

mijiga:
@androidmaven  review

mijiga:
> > @androidmaven review
> 
> Looking good Chief, but one last thing as a cleanup. Can you run optimise code imports so the lingering `import com.uxcam.UXCam` in the fragments are being removed?
> 
> Thanks.

Done

androidmaven:
Thanks! Currently doing functional testing. I will revert in a bit.

androidmaven (approved):
LGTM

PR #484, opened 2022-05-19
182237080-accessibility-toggle-bug-fix

androidmaven:
Accessibility intent needed a: Intent.FLAG_ACTIVITY_CLEAR_TOP, instead of FLAG_ACTIVITY_NO_HISTORY


alexosugo (approved):

davkutalek:
@androidmaven nice catch! Please make sure to update the SDK since it is the source of the problem

PR #485, opened 2022-05-19
181796566 bug fix waiting for sms

androidmaven:
Requires pulling the latest `sdk stax` version.


alexosugo (changes requested):
The SDK database version was not bumped up. App crashes on launch.

alexosugo (approved):

PR #521, opened 2022-06-28
minor ui updates

alexosugo:
Mainly suble ui updates like colors and padding. Attempts to flatten the entire UI.


androidmaven (approved):
Looks good

PR #522, opened 2022-06-28
fixes financial tips back navigation

alexosugo:
- fixes issues with back navigation from system button


androidmaven (approved):
Looks good to me.

PR #530, opened 2022-07-05
Home improvements

alexosugo:
- Adds minor styling updates to compose components. 
- Firestore fetches now run from coroutines; makes it easier to manage state.
- Checks for scheduled tip before displaying on home screen. 
- Fixes unboxed usages of `StateFlow`.


androidmaven (approved):
This is an improvement. Thanks for this, it looks good.

PR #546, opened 2022-07-18
182409679 sim feature

androidmaven:
- Sim page with compose.
- New codes use the data-domain-presentation architecture
- Auto update channels db from json file if required after app updates.

N.B
This PR does not yet include: showing actual sim first before displaying it's account information. It currently functions as it is previously, however with code updates to use the new architecture. Hat tip to @alexosugo's suggestion.

unsupported_sim_single

https://user-images.githubusercontent.com/21104847/187464000-213e2685-632f-4b2f-9e92-a29820e33a98.mp4


davkutalek:
@androidmaven There shouldn't be any add flow for SIMs. Once we have the phone permission we automatically show all the users SIM cards.

app/src/main/java/com/hover/stax/accounts/AccountsViewModel.kt
```diff
@@ -2,26 +2,30 @@ package com.hover.stax.accounts
...
+import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.flow.receiveAsFlow
 import kotlinx.coroutines.launch
 
-class AccountsViewModel(application: Application, val repo: AccountRepo, val actionRepo: ActionRepo, private val bonusRepo: BonusRepo) : AndroidViewModel(application),
+class AccountsViewModel(application: Application,
```
alexosugo, line 24: All Sim functions should move to their own `ViewModel`. We're moving from function-based ViewModels to screen-based ViewModels. It's easier to track actions and side-effects that way. Use use-cases to compose your functionality as they are reusable.

app/src/main/java/com/hover/stax/addChannels/ChannelsViewModel.kt
```diff
@@ -31,12 +32,17 @@ import kotlinx.coroutines.flow.asSharedFlow
 import kotlinx.coroutines.flow.receiveAsFlow
 import kotlinx.coroutines.launch
 import org.json.JSONObject
+import timber.log.Timber
 
-class ChannelsViewModel(application: Application, val repo: ChannelRepo, val accountRepo: AccountRepo, val actionRepo: ActionRepo, private val bonusRepo: BonusRepo) : AndroidViewModel(application),
+class ChannelsViewModel(application: Application, val repo: ChannelRepo,
```
alexosugo, line 37: Please refer to the above comment.

app/src/main/java/com/hover/stax/bonus/BonusViewModel.kt
```diff
@@ -17,7 +18,7 @@ import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
 import timber.log.Timber
 
-class BonusViewModel(val repo: BonusRepo, private val channelRepo: ChannelRepo) : ViewModel() {
+class BonusViewModel(val repo: BonusRepo, private val channelRepo: ChannelRepo,private val presentSimUseCase: GetPresentSimUseCase) : ViewModel() {
```
alexosugo, line 21: This class will eventually be removed in favor of the `GetBonusUseCase`. Make the changes there as well. Eventually, fetch bonus usages will be directed to the usecase.

app/src/main/java/com/hover/stax/data/repository/AccountRepositoryImpl.kt
```diff
@@ -22,30 +24,34 @@ import org.json.JSONObject
 import org.koin.core.component.KoinComponent
 import org.koin.core.component.inject
 
-class AccountRepositoryImpl(val accountRepo: AccountRepo, val channelRepo: ChannelRepo, val actionRepo: ActionRepo, private val coroutineDispatcher: CoroutineDispatcher) : AccountRepository, PushNotificationTopicsInterface, KoinComponent {
+class AccountRepositoryImpl(val accountRepo: AccountRepo, val channelRepo: ChannelRepo, private val presentSimUseCase: GetPresentSimUseCase,  val actionRepo: ActionRepo, private val coroutineDispatcher: CoroutineDispatcher) : AccountRepository, PushNotificationTopicsInterface, KoinComponent {
```
alexosugo, line 27: You should use the repo not the usecase here. UseCases should be used only on the presentation layer. While testing, we provide mocks of the repos used.

app/src/main/java/com/hover/stax/data/repository/BonusRepositoryImpl.kt
```diff
@@ -16,7 +17,9 @@ import kotlinx.coroutines.launch
 import kotlinx.coroutines.tasks.await
 import kotlinx.coroutines.withContext
 
-class BonusRepositoryImpl(private val bonusRepo: BonusRepo, private val channelRepo: ChannelRepo, private val coroutineDispatcher: CoroutineDispatcher) : BonusRepository {
+class BonusRepositoryImpl(private val bonusRepo: BonusRepo,
+                          private val presentSimUseCase: GetPresentSimUseCase,
```
alexosugo, line 21: Refer to above comment.

app/src/main/java/com/hover/stax/presentation/simcard/SimFragment.kt
```diff
@@ -0,0 +1,127 @@
...
+import com.hover.stax.utils.UIHelper
+import com.hover.stax.utils.collectLifecycleFlow
+import org.koin.androidx.viewmodel.ext.android.sharedViewModel
+import timber.log.Timber
+
+class SimFragment : Fragment(), BalanceTapListener {
+
+	private var _binding: FragmentSimBinding? = null
+	private val binding get() = _binding!!
+	private val accountViewModel: AccountsViewModel by sharedViewModel()
```
alexosugo, line 31: SimFragment should only use one viewmodel.

alexosugo (changes requested):
Clicking continue on the dialog that shows up after clicking the link sim button does nothing. I'm not sure if this is a known bug.

app/src/main/java/com/hover/stax/accounts/AccountsViewModel.kt
androidmaven: Thanks for this. Code updated

app/src/main/java/com/hover/stax/addChannels/ChannelsViewModel.kt
androidmaven: Thanks!

app/src/main/java/com/hover/stax/bonus/BonusViewModel.kt
androidmaven: Code updated.

app/src/main/java/com/hover/stax/data/repository/AccountRepositoryImpl.kt
androidmaven: Thanks for this.

app/src/main/java/com/hover/stax/presentation/simcard/SimFragment.kt
androidmaven: Code updated.

app/src/main/java/com/hover/stax/domain/use_case/sims/GetLivePresentSimUseCase.kt
```diff
@@ -0,0 +1,10 @@
+package com.hover.stax.domain.use_case.sims
+import androidx.lifecycle.LiveData
+import com.hover.sdk.sims.SimInfo
+import com.hover.stax.domain.repository.SimRepository
+
+class GetLivePresentSimUseCase(private val simRepository: SimRepository) {
```
alexosugo, line 6: This can be combined with the `GetPresentSimsUseCase` to form a single class, say `GetPresentSims` since they both do the same thing.

app/src/main/java/com/hover/stax/domain/use_case/sims/GetPresentSimUseCase.kt
```diff
@@ -0,0 +1,13 @@
+package com.hover.stax.domain.use_case.sims
+import com.hover.sdk.sims.SimInfo
+import com.hover.stax.domain.model.Bounty
+import com.hover.stax.domain.repository.BountyRepository
+import com.hover.stax.domain.repository.SimRepository
+
+class GetPresentSimUseCase(private val simRepository: SimRepository, private val bountyRepository: BountyRepository) {
```
alexosugo, line 7: Please refer to the previous comment.

app/src/main/java/com/hover/stax/presentation/simcard/SimViewModel.kt
```diff
@@ -0,0 +1,36 @@
...
+		telecomAccounts.addSource(simSubscriptionIds, this::getTelecomAccounts)
+	}
+
+	private fun setSubscriptionIds(sims: List<SimInfo>) {
+		simSubscriptionIds.postValue(sims.map { it.subscriptionId }.toIntArray())
+	}
+
+	private fun getTelecomAccounts(subIds: IntArray){
+		viewModelScope.launch(Dispatchers.IO) {
+			telecomAccounts.postValue(repo.getTelecomAccounts(subIds))
```
alexosugo, line 32: Use the `GetAccountsUseCase` to fetch these accounts. That way, you can abstract away the `AccountRepo`.

alexosugo (changes requested):
After clicking the `Link Sim Cards` button, I get the confirmation dialog but nothing happens once I approve it.

davkutalek:
@androidmaven my phone always shows the "Link your SIM cards to Stax" even tho I've granted the permissions

app/src/main/java/com/hover/stax/data/repository/AccountRepositoryImpl.kt
```diff
@@ -28,22 +30,25 @@ class AccountRepositoryImpl(val accountRepo: AccountRepo, val channelRepo: Chann
...
-            val accountName: String = if (getFetchAccountAction(channel.id) == null) channel.name else channel.name.plus(PLACEHOLDER )//placeholder alias for easier identification later
-            Account(
-                accountName, channel.name, channel.logoUrl, channel.accountNo, channel.id, channel.countryAlpha2,
-                channel.id, channel.primaryColorHex, channel.secondaryColorHex, defaultAccount == null && index == 0
-            )
-        }.onEach {
-            logChoice(it)
-            ActionApi.scheduleActionConfigUpdate(it.countryAlpha2, 24, context)
-        }
+    override suspend fun createAccount(channel: Channel, subscriptionId: Int, isDefault: Boolean) {
```
alexosugo, line 39: Things to note on this function:
- The account name allocation needs to be changed to match the one in `ChannelsViewModel`.
- Generally, batch inserts are faster than multiple single inserts. 
- Batch inserts are atomic, meaning if one insert/update fails, the entire transaction is rolled back, depending on use case, this may be a desired effect.

app/src/main/java/com/hover/stax/data/repository/ChannelRepositoryImpl.kt
```diff
@@ -4,15 +4,15 @@ import com.hover.sdk.actions.HoverAction
...
 import com.hover.stax.domain.repository.ChannelRepository
 
 private const val MAX_LOOKUP_COUNT = 40
 
-class ChannelRepositoryImpl(val channelRepo: ChannelRepo) : ChannelRepository {
+class ChannelRepositoryImpl(val channelRepo: ChannelRepo,val simRepo: SimRepo) : ChannelRepository {
 
-    override val presentSims: List<SimInfo>
-        get() = channelRepo.presentSims
+    override suspend fun presentSims(): List<SimInfo> = simRepo.getPresentSims()
```
alexosugo, line 15: Duplicate method as the one in `SimRepositoryImpl`.

app/src/main/java/com/hover/stax/home/MainActivity.kt
```diff
@@ -78,6 +84,17 @@ class MainActivity : AbstractGoogleAuthActivity(), BiometricChecker.AuthListener
             createFromRequest(intent.getStringExtra(REQUEST_LINK)!!)
         }
     }
+    private fun observerSims() {
```
alexosugo, line 87: Consider using `StateFlow`. It makes doing this unnecessary.

app/src/main/java/com/hover/stax/presentation/simcard/SimScreen.kt
```diff
@@ -0,0 +1,333 @@
...
+@Composable
+fun SimScreen(simScreenClickFunctions: SimScreenClickFunctions,
+              balanceTapListener: BalanceTapListener) {
+	val simViewModel : SimViewModel = getViewModel()
+	val bonusViewModel: BonusViewModel = getViewModel()
+
+	val accounts = simViewModel.telecomAccounts.observeAsState(initial = null)
+	val hasNetwork by NetworkMonitor.StateLiveData.get().observeAsState(initial = false)
+	val presentSims = simViewModel.presentSims.observeAsState(initial = emptyList())
+	val bonuses = bonusViewModel.bonusList.collectAsState()
```
alexosugo, line 55: There exists a usecase to fetch bonuses. Add it to `SimViewModel` and remove the `BonusViewModel`.

alexosugo (changes requested):
It is functioning well now. A few things to consider:
- The `check balance` button has a blue background around it (other than the button color). Try increasing the brightness on your device to see it. 
- For all queries to the db, opt for `Flow` instead of `LiveData` as they observe database changes and propagate them as they occur. 
- For all viewmodels to be used in compose components, use `StateFlow` and `SharedFlow`. They remove the need to add empty observers and we remove the risk of the `multiple observers on same thread` bug.

alexosugo (approved):

alexosugo:
@davkutalek Awaiting your final review. The feature should be functioning as intended now.

davkutalek (changes requested):
Overall this is great. There are some basic refactoring to do in terms of de-duplicating the code for account creation. 

I had a couple other issues:
- When updating from my existing install it didn't add my accounts. It asked for the permissions (even though they were granted) and then didn't do anything.
- The permission ask is always at the top even when my SIM is detected and shown.
- All detected SIMs should be shown, but if they have no channel should just say "unsupported". This way it is clear that we have detected it.

app/src/main/java/com/hover/stax/presentation/sim/SimScreen.kt
```diff
@@ -0,0 +1,477 @@
...
+            simInfo.operatorName ?: "Null",
+            simInfo.networkOperator ?: "Null",
+            simInfo.countryIso ?: "Null")
+    }
+
+    //Sometimes operator name comes in as : No service,
+    //We don't want user to see: "We detected a No service SIM in slot 1..."
+    //Instead it should be: "We detected a SIM in Slot 1", otherwise show "We detected a MTN SIM in Slot 1"
+    val actualOperatorName = simInfo?.operatorName ?: ""
+    val displayedNetworkName = if((actualOperatorName).contains("No service")) "" else actualOperatorName
```
davkutalek, line 258: It is strange that operatorName is empty, that is the correct variable. I have reproduced the issue on my device/SIM will look into how to fix it. This is fine in the meantime, just want some minor UI changes which I will describe.

davkutalek (approved):
@androidmaven this is great, just needs those minor UI changes as mentioned, see below. I will do some work on the SDK to fix the issue with the name, not a blocker for release.

UI changes:
- Change "Your linked SIM cards" to "Your SIM cards" and make it the title of the page (replace the "SIM" title above)
- For SIMs that aren't supported let's use the same layout with the telco name and slot number and icon at the top. I will find a way to make sure there is always a name, but if it is blank just put "Unknown" for now. The icon can be blank for now, but there is supposedly a way to get it from Android, I will work on this as well. In place of the check balance button just put the word unsupported. Then remove the first paragraph and the full text  can read "If this SIM supports USSD please reach out to our support team so we can integrate it." The email button should be the style of the buy airtime button (not blue).

davkutalek:
@androidmaven looks good, only minor thing from me is that Unsupported should just be text not a button

PR #550, opened 2022-07-25
182766412 report balance check error

androidmaven:
- Updated naming to: flashAndReportMessage and flashAndReportError.
- Make all toast logged to analytics.
- Log toast error notifications to Crashlytics.

### DIFF
--- a/app/src/main/java/com/hover/stax/utils/UIHelper.kt
+++ b/app/src/main/java/com/hover/stax/utils/UIHelper.kt
@@ -34,7 +34,7 @@ object UIHelper {
 
     private const val INITIAL_ITEMS_FETCH = 30
 
-    fun flashMessage(context: Context, view: View?, message: String?) {
+    fun flashMessage(context: Context, view: View?, message: String) {
         if (view == null) flashMessage(context, message) else showSnack(view, message)
     }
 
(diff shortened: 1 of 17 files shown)


davkutalek (approved):
LGTM

app/src/main/java/com/hover/stax/utils/UIHelper.kt
```diff
@@ -44,12 +44,17 @@ object UIHelper {
...
-    fun flashMessage(context: Context, message: String?) {
+    fun flashMessage(context: Context, message: String, isAppError: Boolean = false) {
         Toast.makeText(context, message, Toast.LENGTH_SHORT).show()
+        AnalyticsUtil.logAnalyticsEvent(message, context)
+        if(isAppError) AnalyticsUtil.logErrorAndReportToFirebase(context.getString(R.string.toast_err_tag), message, null)
     }
 
-    fun flashMessage(context: Context, messageRes: Int) {
-        Toast.makeText(context, context.getString(messageRes), Toast.LENGTH_SHORT).show()
+    fun flashMessage(context: Context, messageRes: Int, isAppError: Boolean = false) {
```
alexosugo, line 53: Doing this changes what the method is meant to do. Let's separate logging from showing toasts. In keeping with SOLID principles, it'll help during testing.

alexosugo (changes requested):

app/src/main/java/com/hover/stax/utils/UIHelper.kt
davkutalek: @alexosugo I partly disagree. We always want flash messages to be logged, so the analytics call should be here. The method name should change to make this clear `flashAndReportMessage` or something, but we specifically don't want to have to remember to log the message every time we show a flash. 

That said the boolean arg is a clear violation of SOLID, so prob want a second method instead for errors `flashAndReportError` or something.

alexosugo: @davkutalek Agreed. Having a separate method makes sense.

alexosugo (approved):

PR #555, opened 2022-07-29
Rewording get_rewarded text value

wachukxs:
Rewording text value for get_rewarded from "Get Rewarded" to "Get Airtime Bonus". This is a more nudging CTA.


androidmaven (approved):
Thanks!

PR #558, opened 2022-08-09
182744774 make financial tips dismissible from homescreen

androidmaven:
When the user clicks on the financial-tip close button, it dismisses that tip from home screen throughout the day. 

When visiting the home screen another next day, it shows the new tip, until the user dismisses it.


alexosugo (approved):
Works great. Please remember to revert the date condition before merging the changes.

PR #571, opened 2022-08-29
183012754 optional login

alexosugo:
- Refactored login code to use new endpoints for staging. 
- Uses Retrofit as the main network client. 
- Changes login code to adhere to new app architecture. 
- Adds Login with Google on the Welcome screen. 
- Minor refactors of Welcome Fragment.


app/src/main/res/values-om/strings.xml
```diff
@@ -5,10 +5,10 @@
     <string name="app_name" translatable="false">Stax</string>
     <string name="payment_root_url" translatable="false">https://stax.me/pay?%1$s</string>
     <string name="payment_url_end" translatable="false">%1$s-%2$d-%3$s-%4$d</string>
-    <string name="root_url" translatable="false">https://www.usehover.com</string>
```
davkutalek, line 8: These strings should not be duplicated in localization files, that is the point of the translatable flag

davkutalek (approved):
LGTM, just minor issue with the strings

androidmaven (approved):
LGTM.

Thanks!

app/src/main/res/values-om/strings.xml
alexosugo: @davkutalek Yeah, I agree. I ran into them while working on this feature and removed some of them. I'll clear the rest before merging.

PR #572, opened 2022-08-30
182416143 verified mappers

davkutalek:
Changes bounty instructions to tell users to apply. Once they are marked as a verified mapper on the backend then they should be able to load the bounty list screen.

@androidmaven @alexosugo having 2 minor issues before this is ready: 1) CDATA isn't working for the instructions, not sure why.  2) User data needs to be updated in the background to update the isMapper value. This should already be part of the rewards PR, but it looks like that PR has not been opened yet.


alexosugo:
@davkutalek I've pushed a commit that fixes the styling issues for the `BountyApplicationFragment`. CDATA text doesn't work if set directly in XML. You have to format it in your fragment and set it there. I've also added some line breaks to the instructions; double for now but you can edit as desired.

PR #577, opened 2022-09-02
183152718 fix bonus display bug

androidmaven:
For Bonus
- Added sim hni's to the parameters of bonus on Firebase.
- Added sim Hnis to bonus model.
- Deleted BonusViewModel - I don't think we need this any longer considering the new architecture. Bonuses are now retrieved via the respective screen's ViewModel and the business logic is handled by the BonusRepositoryImpl class.
- Some minor refactor to the BonusRepositoryImpl itself.

For SimScreen:
Each SIM item is now smarter to display the correct bonus amounts or none.


davkutalek (approved):
LGTM. I do have an alternate idea for dealing with actions from other channels and including information about the bonus directly in the action config, but it will take some time to get to that. For now this seems like a good improvement.

PR #579, opened 2022-09-07
added institution name to the transaction history item

androidmaven:
Added institution-type information to the transaction history so that where it previously says

- Waiting for confirmation now becomes "Waiting for {type} confirmation".
- New Balance now becomes "New {type} Balance".

a_improved_history_item


davkutalek (approved):
LGTM, let's just use institution name instead of type

PR #580, opened 2022-09-07
fixed incorrect balance amount

androidmaven:


app/src/main/java/com/hover/stax/utils/Utils.kt
```diff
@@ -122,7 +122,7 @@ object Utils {
     fun formatAmount(number: Double?): String {
         return try {
             val formatter = DecimalFormat("#,##0.00")
-            formatter.maximumFractionDigits = 0
+            formatter.maximumFractionDigits = 3
```
davkutalek, line 125: Should be 2 not 3

app/src/test/java/com/hover/stax/utils/UtilsTest.kt
```diff
@@ -0,0 +1,13 @@
+package com.hover.stax.utils
+
+import org.junit.Assert
+import org.junit.Test
+
+
+class UtilsTest {
+	@Test
```
davkutalek, line 8: Should have a couple of tests to check what happens with eg `43`, `43.1` and `43.986` since it is possible to get any of these back from various USSD services.

davkutalek (changes requested):
Minor changes requested, see comments

app/src/test/java/com/hover/stax/utils/UtilsTest.kt
davkutalek: Ideally the different inputs should be separate tests so that you can name them and anyone can know what you're testing  for

davkutalek (approved):

PR #594, opened 2022-10-05
183284652 crash fixes

davkutalek:
Fixes 2 issues found by John https://www.pivotaltracker.com/story/show/183284652

the change to make the institution types non-null caused issues because you tried to add columns that already existed. There are SQL commands to change a column from nullable to not, but I think in this case they are unnecessary because you've already set a default value for the column. There should be no way to make the column null if it has a default (and we aren't sending null from our server). Let me know if you think there is an issue with this approach.


androidmaven:
Thanks for fixing this, the bug was my mistake

PR #599, opened 2022-10-11
183505442 db migration fixes for 42.json

androidmaven:
The integrity of 42json changed in development from what it used to be in v1.12.16, it ought to be the same all through.

- I pasted the original content and a few simple refactoring in this fix.
- Added manual migration to drop and recreate channels table so that other tables persist.

I'd be more careful with them in the future.

### DIFF
--- a/app/src/main/java/com/hover/stax/database/AppDatabase.kt
+++ b/app/src/main/java/com/hover/stax/database/AppDatabase.kt
@@ -214,17 +214,11 @@ abstract class AppDatabase : RoomDatabase() {
             database.execSQL("CREATE UNIQUE INDEX IF NOT EXISTS `index_stax_transactions_uuid` ON `stax_transactions` (`uuid`)")
         }
 
-        private val M42_43 = Migration(42, 43) { database -> //accounts table changes
-            database.execSQL("ALTER TABLE channels ADD COLUMN institution_type TEXT NOT NULL DEFAULT 'bank'")
-            database.execSQL("ALTER TABLE accounts ADD COLUMN institution_type TEXT NOT NULL DEFAULT 'bank'")
-            database.execSQL("ALTER TABLE accounts ADD COLUMN sim_subscription_id INTEGER NOT NULL DEFAULT -1")
-        }
-
-
-
-
-        /* When using fallbackToDestructiveMigrationFrom
+        /* By using fallbackToDestructiveMigrationFrom for v42,
+        It drops the V42 table, and creates a new one for v43, hence, no need for migration.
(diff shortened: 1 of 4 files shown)


davkutalek:
I think I want to try dropping just the channel table since otherwise users will have to re-add their accounts and that is a bad xp/we don't have a process for asking them to do that.

davkutalek:
@androidmaven Nice, this works for me tho I haven't tested various upgrade paths yet.

davkutalek:
@androidmaven One thing to note is there are bunch of old fields on channels that should be removed: selected, defaultAccount, pin, latestBalance, latestBalanceTimestamp, accountNo. Now would prob be the best time to get rid of them, but when I try I do get errors.

davkutalek:
I also think doing the work in 42-43 may cause issues since some users may have a new db version. Haven't confirmed yet

app/src/main/java/com/hover/stax/database/AppDatabase.kt
```diff
@@ -215,10 +215,25 @@ abstract class AppDatabase : RoomDatabase() {
...
+
+            database.execSQL(
+                "INSERT INTO channels_new (id, name, country_alpha2, root_code, currency, hni_list, logo_url, institution_id, primary_color_hex, published, secondary_color_hex, selected, defaultAccount, isFavorite, pin, latestBalance, latestBalanceTimestamp, account_no)" +
+                        " SELECT id, name, country_alpha2, root_code, currency, hni_list, logo_url, institution_id, primary_color_hex, published, secondary_color_hex, selected, defaultAccount, isFavorite, pin, latestBalance, latestBalanceTimestamp, account_no FROM channels"
+            )
+            database.execSQL("DROP TABLE channels")
+            database.execSQL("ALTER TABLE channels_new RENAME TO channels")
+
+
+            database.execSQL("ALTER TABLE accounts ADD COLUMN institution_type TEXT NOT NULL DEFAULT 'bank'")
```
davkutalek, line 233: I believe this line may also cause errors, which may require us to drop and re-create the accounts table

androidmaven: I don't think this causes an error; it remained the same after double-checking the JSON files before and after the divergent.

androidmaven: For the removal of selected, default account, pin, latest balance, latestBalanceTimestamp, accountNo. 

I'd suggest we do that for newer database in a separate story.

davkutalek (approved):
Tested every migration path I could, seems to work for all of them

PR #592, opened 2022-09-19
Sim and component improvements

davkutalek:
Refactors the Sim use case to separate concerns better. Refactors UI components into more generic components we will use elsewhere. Will add further comments later this week.

PR #605, opened 2022-10-12
183477630 same sim type bugfix

androidmaven:
- Simplified SimViewModel

### DIFF
--- a/app/src/main/java/com/hover/stax/data/repository/AccountRepositoryImpl.kt
+++ b/app/src/main/java/com/hover/stax/data/repository/AccountRepositoryImpl.kt
@@ -80,9 +80,8 @@ class AccountRepositoryImpl(val accountRepo: AccountRepo, val channelRepo: Chann
         }
     }
 
-    override fun getTelecomAccounts(simSubscriptionIds: IntArray): Flow<List<Account>> {
-        return accountRepo.
-        getTelecomAccounts(simSubscriptionIds)
+    override fun getTelecomAccounts(simSubscriptionIds: IntArray): List<Account> {
+        return accountRepo.getTelecomAccounts(simSubscriptionIds)
     }
 
     private fun getFetchAccountAction(channelId: Int): HoverAction? = actionRepo.getActions(channelId, HoverAction.FETCH_ACCOUNTS).firstOrNull()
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/hover/stax/data/repository/AccountRepositoryImpl.kt
```diff
@@ -40,13 +40,13 @@ class AccountRepositoryImpl(val accountRepo: AccountRepo, val channelRepo: Chann
     override suspend fun createAccount(channel: Channel, simSubscriptionId: Int, isDefault: Boolean) {
         val accountName: String = if (getFetchAccountAction(channel.id) == null) channel.name else PLACEHOLDER //placeholder alias for easier identification later
         val account = Account(
-            accountName, channel.name, channel.logoUrl, channel.accountNo, channel.id, channel.institutionType, channel.countryAlpha2,
```
davkutalek, line 43: Why change the id to the institution? That is certainly going to break something.

Also, why is this using the default constructor instead of a constructor that you pass the channel to?

davkutalek (commented):
Not sure what this actually fixes. Would rather not make arbitrary changes when we are trying to keep things stable.

app/src/main/java/com/hover/stax/data/repository/AccountRepositoryImpl.kt
androidmaven: I looked at the Account's model the parameter expects institutionId, where it was previously passed channelId.

I can change back, but I suppose we need to attend it again.

androidmaven:
@davkutalek, at the moment, this primarily refactors simViewModel. We were discussing on Slack what to do with the account name ish.

app/src/main/java/com/hover/stax/data/repository/AccountRepositoryImpl.kt
davkutalek: ah I see. But this points to why I will keep bringing up the constructor: there is already a constructor that takes channel as an argument and deals with the details correctly. Why aren't you using that one?

davkutalek:
This simviewmodel refactor isn't really doing much tho... It appears to be recreating/updating the account each time, which I don't think is what we want. Ultimately the view model should not be intermediated in the process of finding and creating the accounts since that is business logic. This is what usecases are for. See #592 ListSimsUseCase and SimViewModel. That PR is not quite ready, but I think we can get it ready by next week. In the meantime let's just solve the immediate issue with the names

app/src/main/java/com/hover/stax/database/AppDatabase.kt
```diff
@@ -217,13 +217,16 @@ abstract class AppDatabase : RoomDatabase() {
         private val M42_43 = Migration(42, 43) { database -> //accounts table changes
             database.execSQL("ALTER TABLE accounts ADD COLUMN sim_subscription_id INTEGER NOT NULL DEFAULT -1")
         }
+        private val M45_46 = Migration(45, 46) { database -> //accounts table indices changes
+            database.execSQL("CREATE UNIQUE INDEX index_accounts_name_sim_subscription_id ON accounts (name, sim_subscription_id)")
```
davkutalek, line 221: need to drop the old index first

PR #616, opened 2022-10-13
Depreciate channel fields

davkutalek:
There were a number of fields that predate the creation of Accounts. Some of them have been used because the comment saying not to use them is ignored (since that is bad practice anyway). The only user discernible effect of this removal is that we still show selected accounts in the add accounts list. This is how it should be, as users could have multiple accounts connected to different SIM cards. It does still prevent adding a second account however since duplicate account names are not yet allowed.


davkutalek:
Tested a bunch of upgrade paths, no issues. Merging for release

PR #635, opened 2022-10-18
Update BUG_REPORT.yml

hilderbrandtjohn:


androidmaven:
Thank you for this suggestion, John. I'll let @davkutalek also take a look.

hilderbrandtjohn:
> 

Thank you.I felt that this would necesitate us to categorize the bugs that should be fixed immediately and the ones that can be worked on within the weekly sprints

jshorland (commented):
Adding severity and priority for bug categorization. Thanks @hilderbrandtjohn !

PR #723, opened 2022-10-26
Likely fix for recipient network logo not refreshing/resetting

davkutalek:
I am unable to reproduce the issue. However, it is likely that every time the dropdown was updated a new target was being created, leaving the old one hanging. Now only creates one target per dropdown instance.


androidmaven (approved):
Yes, I think so too.

PR #742, opened 2022-10-27
Remove fetch accounts

davkutalek:
Remove use of fetch accounts actions, remove use of placeholder as name. Update account db/model fields to make them clearer, fix to the way creating multiple accounts for a given bank or channel works.

Now when an account is first added the check balance action should always be run. If an account name choice is encountered then 1 will always be chosen. Once the transaction has finished Stax will parse any `userAccountList` in the output variables and update/create accounts for each item in the list.

I can't actually test this as I don't have access to any bank USSD channels. Note that it requires latest SDK update.

### DIFF
--- a/app/src/main/java/com/hover/stax/database/AppDatabase.kt
+++ b/app/src/main/java/com/hover/stax/database/AppDatabase.kt
@@ -33,14 +33,15 @@ import java.util.concurrent.Executors
     entities = [
         Channel::class, StaxTransaction::class, StaxContact::class, Request::class, Schedule::class, Account::class, Paybill::class, Merchant::class, StaxUser::class, Bonus::class
     ],
-    version = 46,
+    version = 47,
     autoMigrations = [
         AutoMigration(from = 36, to = 37),
         AutoMigration(from = 37, to = 38),
         AutoMigration(from = 38, to = 39),
         AutoMigration(from = 40, to = 41),
         AutoMigration(from = 41, to = 42),
-        AutoMigration(from = 43, to = 44)
+        AutoMigration(from = 43, to = 44),
+        AutoMigration(from = 46, to = 47)
     ]
 )
 
@@ -232,7 +233,7 @@ abstract class AppDatabase : RoomDatabase() {
 
         private val M44_45: Migration = Migration(44, 45) {}
 
-        private val M45_46 = Migration(45, 46) { database -> //accounts table changes
+        private val M45_46 = Migration(45, 46) { database ->
             database.execSQL(
                 "CREATE TABLE IF NOT EXISTS `channels_new` (`id` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, `name` TEXT NOT NULL, `country_alpha2` TEXT NOT NULL, `root_code` TEXT, " +
                         "`currency` TEXT NOT NULL, `hni_list` TEXT NOT NULL, `logo_url` TEXT NOT NULL, `institution_id` INTEGER NOT NULL, `primary_color_hex` TEXT NOT NULL, `published` INTEGER NOT NULL DEFAULT 0," +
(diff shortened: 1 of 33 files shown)


app/src/main/java/com/hover/stax/hover/AbstractHoverCallerActivity.kt
```diff
@@ -46,7 +46,16 @@ abstract class AbstractHoverCallerActivity : AppCompatActivity(), PushNotificati
...
 
     fun runSession(account: Account, action: HoverAction) {
-        runSession(account, action, null, account.id) // Constants.REQUEST_REQUEST
+        runSession(account, action, null, account.id)
+    }
+
+    private fun getAccountNameExtra(account: Account, action: HoverAction): HashMap<String, String>? {
+        if (action.requiredParams.contains(ACCOUNT_NAME)) {
+            val extras = HashMap<String, String>()
+            extras[ACCOUNT_NAME] =
```
androidmaven, line 55: "extras[ACCOUNT_NAME] =" forgot to set a value.

app/src/main/java/com/hover/stax/presentation/welcome/WelcomeFragment.kt
```diff
@@ -27,7 +27,7 @@ class WelcomeFragment : Fragment() {
             id = R.id.welcomeFragment
             val buttonText = getString(R.string.explore_btn_text)
             setContent {
-                WelcomeScreen(buttonText, { onClickGetStarted() }, { onClickLogin() }, showExploreButton = false)
+                WelcomeScreen(buttonText, { onClickGetStarted() }, { onClickLogin() }, showExploreButton = true)
```
androidmaven, line 30: We are bringing back the explore button, yes?

app/src/main/java/com/hover/stax/requests/NewRequestFragment.kt
```diff
@@ -226,12 +226,12 @@ class NewRequestFragment : AbstractFormFragment(), PushNotificationTopicsInterfa
...
-                payWithDropdown.setState(getString(R.string.incomplete_account_setup_header), AbstractStatefulInput.ERROR)
-                return false
-            }
-        }
+//        requestViewModel.activeAccount.value?.let {
+//            if (!requestViewModel.isValidAccount()) {
+//                payWithDropdown.setState(getString(R.string.incomplete_account_setup_header), AbstractStatefulInput.ERROR)
+//                return false
+//            }
+//        }
```
androidmaven, line 234: Could you please add a note on why it is being commented out?

app/src/main/java/com/hover/stax/database/AppDatabase.kt
```diff
@@ -247,11 +248,16 @@ abstract class AppDatabase : RoomDatabase() {
             database.execSQL("ALTER TABLE channels_new RENAME TO channels")
         }
 
+        private val M47_48 = Migration(47, 48) { database ->
+            database.execSQL("DROP INDEX index_accounts_name")
+            database.execSQL("CREATE UNIQUE INDEX IF NOT EXISTS index_accounts_name_sim_subscription_id ON accounts(name, sim_subscription_id)")
+        }
```
androidmaven, line 254: Gives me joy seeing the correct query. I wrote a couple but didn't work.

androidmaven (changes requested):
This is fantastic, and I left some comments.

app/src/main/java/com/hover/stax/requests/NewRequestFragment.kt
davkutalek: because an account doesn't have to be valid to request money to it

app/src/main/java/com/hover/stax/hover/AbstractHoverCallerActivity.kt
davkutalek: This was meant to be removed, it is actually done in `HoverSession`

app/src/main/java/com/hover/stax/presentation/welcome/WelcomeFragment.kt
davkutalek: yes

PR #764, opened 2022-10-29
183546914 bonus airtime update

davkutalek:
Remove bonus entity and all related classes. Replace with looking up actions directly using the bonus percent field. Also added the prefix required by our bonus airtime channel to the action config so that it is no longer hard-coded.

This also brought to attention a few only sort of related issues:

- Channel country code should not be capitalized since that isn't done for any other entities with country (accounts, actions). Only uppercase when needed for display. 
- Actions should be looked up by whether then can move money *out* of a given account. It doesn't matter what USSD channel they are on. If you can buy airtime with a 3rd party USSD channel then that should show as an option on the buy airtime screen. Therefore use institution + country to look up available actions.


app/src/main/java/com/hover/stax/addChannels/ChannelsViewModel.kt
```diff
@@ -215,25 +210,15 @@ class ChannelsViewModel(application: Application, val repo: ChannelRepo,
...
     private fun updateCountryChannels(channels: List<Channel>?) {
         channels?.let { runFilter(it, filterQuery.value) }
     }
 
     private fun runFilter(channels: List<Channel>, value: String?) {
-        filterBonusChannels(channels.filter { standardizeString(it.toString()).contains(standardizeString(value)) })
-    }
-
-    private fun filterBonusChannels(channels: List<Channel>) = viewModelScope.launch {
-        bonusRepo.collectBonuses.collect { list ->
```
davkutalek, line 230: Bonus channels have nothing to do with this, shouldn't be here

jumaallan (approved):
We need to stop using the `!!` soon..

androidmaven (approved):
Very solid.

davkutalek:
> We need to stop using the `!!` soon..

Fair. The issue here was that the db column isn't ever null, so it shouldn't be nullable. I'll add a db migration

PR #766, opened 2022-10-30
Introduce Chucker to debug network calls on debug builds

jumaallan:
Introducing Chucker makes it easy to debug network calls, but on debug builds only


androidmaven (approved):
This is excellent work. It increases the robustness of testing network calls.

davkutalek (approved):

PR #767, opened 2022-10-30
Show Google Login Failure Error

jumaallan:
When the `StartActivityForResult` returns a result which is not `RESULT_OK`, the app doesn't do anything. This PR allows the app to show a failure message

Easiest way to test this change is to delete your SHA-1 from Firebase, then build app and see if the error appears


androidmaven (approved):
Looks good to me.

jumaallan:
> @jumaallan please change your tabbing to use tabs instead of spaces so that the diffs aren't really crazy. We use tabs in all our projects.

I had the wrong codestyle set on my IDE 😄 

Will fix this

davkutalek (approved):

PR #768, opened 2022-10-31
183613086 amount validation

androidmaven:
This work depends on this hover [SDK merge request](https://github.com/UseHover/hover-sdk/pull/82).

Logic.
ExtractCustomSteps from the action, get the amount validation regex; then, 
- if it is supplied, prefer it to validate the field. 
- Else, use default validation.

### DIFF
--- a/app/src/main/java/com/hover/stax/actions/ActionSelectViewModel.kt
+++ b/app/src/main/java/com/hover/stax/actions/ActionSelectViewModel.kt
@@ -5,14 +5,18 @@ import android.content.Context
 import androidx.lifecycle.AndroidViewModel
 import androidx.lifecycle.MediatorLiveData
 import androidx.lifecycle.MutableLiveData
-
 import com.hover.sdk.actions.HoverAction
-import com.hover.sdk.actions.HoverAction.*
+import com.hover.sdk.actions.HoverAction.ACCOUNT_KEY
+import com.hover.sdk.actions.HoverAction.AMOUNT_KEY
+import com.hover.sdk.actions.HoverAction.NOTE_KEY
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/hover/stax/actions/ActionSelectViewModel.kt
```diff
@@ -42,6 +46,17 @@ class ActionSelectViewModel(application: Application) : AndroidViewModel(applica
         return if (activeAction.value == null) (getApplication() as Context).getString(R.string.action_fielderror) else null
     }
 
+    fun getAmountValidationRegex(): String? {
```
davkutalek, line 49: Doesn't need to be here

app/src/main/java/com/hover/stax/transfers/TransferFragment.kt
```diff
@@ -277,7 +279,7 @@ class TransferFragment : AbstractFormFragment(), ActionSelect.HighlightListener,
         val accountError = accountsViewModel.errorCheck()
         payWithDropdown.setState(accountError, if (accountError == null) AbstractStatefulInput.SUCCESS else AbstractStatefulInput.ERROR)
 
-        val amountError = transferViewModel.amountErrors()
+        val amountError = transferViewModel.amountErrorsCheck(actionSelectViewModel.getAmountValidationRegex())
```
davkutalek, line 282: Improve consistency by making this api look like the recipient error check. Just pass the selected action itself to the error check method. This also separates concerns better - only the error check method needs to know how to get the validation regex.

davkutalek (approved):
Making minor changes and then merging

PR #802, opened 2022-11-07
183659239 service integration bugfix

androidmaven:
Request Actions to prepopulate; 
At the same instant, the channels import.


davkutalek:
@androidmaven this is missing the `ActionsImportWorker` class

davkutalek (changes requested):
Let's just have the SDK start it's own action import worker, see that PR. This will allow us to provide a simple API for any developer to do the same. The other fixes included here can still be merged once RoutingActivity changes are removed.

davkutalek:
@androidmaven I just saw that channelimportworker is the same way with 2 asset files. Please fix that as well.

davkutalek (approved):
LGTM

PR #804, opened 2022-11-09
Fix non standard variables not able to get focus

davkutalek:
I was not able to figure out the root cause of this issue. To fix it, adds the edit boxes in fragment manually, an adapter isn't really necessary anyway. I did try to change it from a recycler view to list view, but the problem was the same.

[Finishes #183687048]


androidmaven (approved):
LGTM

PR #823, opened 2022-11-14
183789969 sim screen hotfix

androidmaven:
Refactored how the bonus is retrieved on SimItem.

Considering that the bonus is default zero on HoverAction, I updated the conditional logic.


jumaallan (approved):

davkutalek (approved):
LGTM

PR #838, opened 2022-11-23
Cleanup Database and Add Database Migration Tests

jumaallan:
This PR
* Abstracts the DAO implementations (Still not perfectly done since some refactoring is needed on how we inject database/handle write)
* Introduce tests for database migrations


androidmaven (approved):
Juma, you are killing it!

PR #879, opened 2022-12-02
183935447 Fix Null institutionId in Accounts

jumaallan:
This PR queries the `channels` table, by checking the `name` and `alias`, then updates the `institution_id` inside the `accounts table`

 I also added a default value  as a back up option (to prevent future crashes)

Finishes #183935447

### DIFF
--- a/app/src/main/java/com/hover/stax/database/Migrations.kt
+++ b/app/src/main/java/com/hover/stax/database/Migrations.kt
@@ -220,5 +220,16 @@ class Migrations {
             database.execSQL("CREATE UNIQUE INDEX IF NOT EXISTS `index_accounts_name_sim_subscription_id` ON `accounts` (`name`, `sim_subscription_id`)")
             database.execSQL("CREATE INDEX IF NOT EXISTS `index_accounts_channelId` ON `accounts` (`channelId`)")
         }
+
+        val M51_52 = Migration(51, 52) { database ->
+            database.execSQL(
+                "INSERT INTO accounts (institutionId)\n" +
+                     "SELECT a.institution_id\n" +
+                     "FROM (SELECT institution_id, name FROM channels) AS a \n" +
+                     "LEFT JOIN accounts AS b \n" +
+                     "ON a.name = b.alias \n" +
+                     "WHERE b.institutionId IS NULL"
+            )
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/hover/stax/database/Migrations.kt
```diff
@@ -220,5 +220,16 @@ class Migrations {
...
             database.execSQL("CREATE INDEX IF NOT EXISTS `index_accounts_channelId` ON `accounts` (`channelId`)")
         }
+
+        val M51_52 = Migration(51, 52) { database ->
+            database.execSQL(
+                "INSERT INTO accounts (institutionId)\n" +
+                     "SELECT a.institution_id\n" +
+                     "FROM (SELECT institution_id, name FROM channels) AS a \n" +
+                     "LEFT JOIN accounts AS b \n" +
+                     "ON a.name = b.alias \n" +
```
davkutalek, line 230: I think you can use the `channel_id` field to do the join, I checked and that field has been present since the beginning. Theoretically name + institution id should work as well, but id is obviously preferable. `alias` is editable by the user.

`"INSERT INTO accounts (institutionId)\n" +
                     "SELECT a.institution_id\n" +
                     "FROM (SELECT institution_id, id FROM channels) AS a \n" +
                     "LEFT JOIN accounts AS b \n" +
                     "ON a.id = b.channel_id \n" +
                     "WHERE b.institutionId IS NULL"`

app/src/main/java/com/hover/stax/database/Migrations.kt
```diff
@@ -220,5 +220,16 @@ class Migrations {
             database.execSQL("CREATE UNIQUE INDEX IF NOT EXISTS `index_accounts_name_sim_subscription_id` ON `accounts` (`name`, `sim_subscription_id`)")
```
davkutalek, line 220: I think the above create/insert calls above may need to be re-done but with the default statement added for `institutionId`?

davkutalek (changes requested):
Minor edit to the insert query. 

I'm a little surprised Room doesn't complain about the default value - there is no actual migration adding it to the DB, but the json schema looks right. I don't know whether that can cause problems or not, I'll do a little testing.

app/src/main/java/com/hover/stax/database/Migrations.kt
jumaallan: That's what I see - no other way to add a default value to it

jumaallan:
> Minor edit to the insert query.
> 
> I'm a little surprised Room doesn't complain about the default value - there is no actual migration adding it to the DB, but the json schema looks right. I don't know whether that can cause problems or not, I'll do a little testing.

This is ready to ship now

I had to recreate the accounts table, but add a default value to the institutionId column - then add another query to backfill it if null or has a default value

davkutalek (approved):
LGTM

PR #880, opened 2022-12-02
183857624 Update Balance on Home Screen

jumaallan:
Still WIP - but I know what the problem is, working on it  

Finishes #183857624


davkutalek:
nice, the Sim screen balances aren't updating either and may have a similar problem

PR #883, opened 2022-12-06
180956049 partial actions

davkutalek:
This allows the user to check the fee for a p2p or buy airtime transaction before making the actual purchase. They can also check the recipient name using the same technique. Updates Hover calls to use ActivityResultContract which is much simpler than needing the same base activity and using onActivityResult with codes for everything.

Requires https://github.com/UseHover/hover-sdk/pull/65

https://user-images.githubusercontent.com/735966/206589382-7e0552f3-27b5-4253-89d9-8e79f61249a7.mp4


settings.gradle.kts
```diff
@@ -31,6 +31,7 @@ dependencyResolutionManagement {
         }
     }
 }
+enableFeaturePreview("VERSION_CATALOGS")
```
jumaallan, line 34: Updating the wrapper version should fix this - will revert

jumaallan: Just set this 

```
distributionUrl=https\://services.gradle.org/distributions/gradle-7.4-bin.zip
```

PR #885, opened 2022-12-14
183857624 Update Balance on Home and Sim Screens

jumaallan:
I had to move from using StateFlow to a MutableLiveData since the composable doesn't recompose whenever state flow is updated. Not sure if this has anything to do with the nested UI, but it looks like the UI is destroyed and recreated when the user manually navigates to home screen using the home tab option, and doesn't change when the view was in the backstack - someone pressing back button 

Finishes #183857624


davkutalek (approved):
I don't really understand the issue with the flow, I'll try to do some more reading/experimenting.

jumaallan:
Has some wierd behaviour


Try see if you find a good solution

PR #890, opened 2022-12-19
fix for formatting retry amount. Removed the trailing .## to adds .00

wachukxs:
I noticed we're still having this issue.

Screenshot 2022-12-19 at 3 57 31 PM

The formatter is using decimals (the ```.##``` part of ```"###0.##"```), which will force ```456``` to be ```456.00``` and that's not what we want.

This PR is fixing that.

### DIFF
--- a/app/src/main/java/com/hover/stax/utils/Utils.kt
+++ b/app/src/main/java/com/hover/stax/utils/Utils.kt
@@ -148,8 +148,6 @@ object Utils {
     fun formatAmountForUSSD(number: Double?): String {
         return try {
             val formatter = DecimalFormat("##0")
-            formatter.maximumFractionDigits = 2
-            formatter.minimumFractionDigits = 0
             formatter.format(number)
         } catch (e: Exception) {
             number.toString()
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/hover/stax/utils/Utils.kt
```diff
@@ -147,9 +147,7 @@ object Utils {
     @JvmStatic
     fun formatAmountForUSSD(number: Double?): String {
         return try {
-            val formatter = DecimalFormat("###0.##")
-            formatter.maximumFractionDigits = 2
-            formatter.minimumFractionDigits = 0
+            val formatter = DecimalFormat("##0")
```
wachukxs, line 150: This rounds up, but I think it's safe, cause all the times this has caused a failed transaction, the value was ```.00```

davkutalek: This isn't right, some services do allow decimal values, so this wont work correctly for those.

davkutalek (commented):
The user still having this problem is on an older version of Stax. The fix was included in 1.18.3, which has not gone to production yet due to Google's issues with Accessibility. In any case the user experiencing the problem is on an older version. The existing code is tested and seems to work correctly.

PR #919, opened 2023-01-24
adding screenshots to errors

wachukxs:


wachukxs:
Apparently we already capture user id in sentry, which I assume is the device ID

davkutalek:
@wachukxs I wouldn't assume that, it may be automatic and different than our device ID. It would be good if they were the same if they aren't

wachukxs:
@davkutalek you're right!! I've updated the PR.

davkutalek (changes requested):
Configuration options like this don't belong in a Fragment, they are normally set in the launch activity or ApplicationInstance. In this case they actually go in the SDK in the `AnalyticsSingleton`. I think the Android Manifest change is fine here cause we don't want screenshots from developer's apps.

wachukxs:
@davkutalek I've reverted the configurations changed I made in the Fragment file. Just need one last review from you on this.

davkutalek (approved):

PR #920, opened 2023-01-28
[260123][Modularisation] Introducing modularisation: Added database module and moved StaxUser database logic

anselmoalexandre:
Hello,
In this PR, I intent to introduce the concept of modularised project in which we can benefit from vast advantages of modularisation such as:
- Code ownership,
- Reduce build times,
- Reusable code, etc

To start this long journey, I moved the smallest part of the app (user related database classes in this case) to database module. If the idea gets accepted, we can work on moving/modularising the hole project.

Other PRs will follow with more improvements

Thanks


app/src/androidTest/java/com/hover/stax/MigrationsTest.kt
```diff
@@ -18,7 +18,6 @@ package com.hover.stax
 import androidx.room.testing.MigrationTestHelper
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.platform.app.InstrumentationRegistry
-import com.hover.stax.database.AppDatabase
```
jumaallan, line 21: This breaks the test - need to import the new import path

database/build.gradle.kts
```diff
@@ -0,0 +1,39 @@
...
+        targetCompatibility = JavaVersion.VERSION_1_8
+    }
+
+    kotlinOptions {
+        jvmTarget = "1.8"
+    }
+}
+
+dependencies {
+    // To Refactor: Expose those dependencies as api implementation
```
jumaallan, line 27: Good point

database/build.gradle.kts
```diff
@@ -0,0 +1,39 @@
...
+    // Room
+    implementation(libs.bundles.room)
+    kapt(libs.room.compiler)
+    androidTestImplementation(libs.room.test)
+
+    // DI
+    implementation(libs.bundles.koin)
+
+    // Tests
+    testImplementation(libs.bundles.test)
```
jumaallan, line 38: We also need a `test` module to `api` into all other modules

database/src/main/java/com/hover/stax/database/StaxDatabase.kt
```diff
@@ -0,0 +1,5 @@
+package com.hover.stax.database
+
+import androidx.room.RoomDatabase
+
+abstract class StaxDatabase : RoomDatabase()
```
jumaallan, line 5: This is not implemented anywhere

Or is it a placeholder to use when we migrate `AppDatabase`?

database/build.gradle.kts
anselmoalexandre: Yap, I added it like that so we can all discuss the modularisation strategy and implement it as per our discussion. This PR is just a kick off

app/src/androidTest/java/com/hover/stax/MigrationsTest.kt
anselmoalexandre: Which tests cases are broken? Will check this

jumaallan (commented):
I noticed that this PR only moves the user dao into the new module - do you want to have subsequent PR? Will that also cover the implementation for `StaxDatabase` from `AppDatabase` ?

One more suggestion I had in mind - do we want to group the features/library modules like this repo for example - https://github.com/android/nowinandroid

we could have


core
* database
* networking etc


features
* feature1
* feature2

app/src/androidTest/java/com/hover/stax/MigrationsTest.kt
jumaallan: This specific test relies on AppDatabase - so it has a missing import here

```
@get:Rule
    val helper: MigrationTestHelper = MigrationTestHelper(
        InstrumentationRegistry.getInstrumentation(), AppDatabase::class.java
    )
```

database/build.gradle.kts
jumaallan: Perfect

database/src/main/java/com/hover/stax/database/StaxDatabase.kt
jumaallan: Great

anselmoalexandre:
> I noticed that this PR only moves the user dao into the new module - do you want to have subsequent PR? Will that also cover the implementation for `StaxDatabase` from `AppDatabase` ?
> 
> We also need to move all the remaining DAOs later as well - same thing you have done with the user setup
> 
> One more suggestion I had in mind - do we want to group the features/library modules like this repo for example - https://github.com/android/nowinandroid
> 
> we could have
> 
> core
> 
> * database
> * networking etc
> 
> features
> 
> * feature1
> * feature2

Yes, that what I had in mind but I had to push something and show where we can start. 
So this PR only moves StaxUser related classes but more PRs will follow migrating the `database`, `networking`, and `features`.

About grouping that would be our choice and I would like to discuss that together.
IMHO grouping is the best, but perhaps I'll need a quick walkthrough so I can see what different type of modules we might have to create, so we can create a better grouping

app/src/androidTest/java/com/hover/stax/MigrationsTest.kt
anselmoalexandre: Alright, will check this

anselmoalexandre:
> I noticed that this PR only moves the user dao into the new module - do you want to have subsequent PR? Will that also cover the implementation for `StaxDatabase` from `AppDatabase` ?
> 
> We also need to move all the remaining DAOs later as well - same thing you have done with the user setup
> 
> One more suggestion I had in mind - do we want to group the features/library modules like this repo for example - https://github.com/android/nowinandroid
> 
> we could have
> 
> core
> 
> * database
> * networking etc
> 
> features
> 
> * feature1
> * feature2

Possible setup could be:
- :app
-:auth(authentication)
- :core
   - :database
   - :network
   - :resources
- :features
   - :profile
   - :transactions
- :utils

We can setup this on another PR (I honestly don't like huge PRs, I prefer to keep the PRs small)

jumaallan:
> 

Perfect

We are setting a slack channel so that should make these discussions faster too

jumaallan (approved):
Looks good to me

jumaallan:
@anselmoalexandre Do you want to fix this conflict and merge this first?

anselmoalexandre:
> @anselmoalexandre Do you want to fix this conflict and merge this first?

No need Principal @jumaallan 
I already merged this PR on the other PR for Modularisation [here](https://github.com/UseHover/Stax/pull/940) 

So we can abandon this one once the other one goes in

jumaallan:
Closing this in favour of #940

PR #940, opened 2023-02-02
[020223] Stax Modular Project setup

anselmoalexandre:
In this PR, I propose the setup for modularised Stax project

Ofc, this modules will have migrated related classes from our current code base
NOTE that, as we migrate, more modules will be added and some refactoring will follow as well

Let me know what you team think
Thanks


anselmoalexandre:
@jumaallan can we polish this branch and merge it asap so we can carry on with the code migration?

anselmoalexandre:
What do you think if we create the `android` block on a separate file and reuse it on all the modules? As I can see, this is boilerplate code right there. What do you think @jumaallan

jumaallan:
> What do you think if we create the `android` block on a separate file and reuse it on all the modules? As I can see, this is boilerplate code right there. What do you think @jumaallan

Have you seen how the `now-in-android` do it? Using `build-logic` which looks really interesting

wangerekaharun:
@anselmoalexandre This looks good. My only comment would be to isolate storage & network modules from the core module and core only have the shared items across the modules

anselmoalexandre:
> @anselmoalexandre This looks good. My only comment would be to isolate storage & network modules from the core module and core only have the shared items across the modules

Thanks @wangerekaharun, but the idea of having network and database into the core is:

- categorised info
- network and database will only store core/base info of the app

If we need something specific we can just implement the specific module using either api or implementation apis.
IMHO this should be there. What do you think @jumaallan

anselmoalexandre:
> > What do you think if we create the `android` block on a separate file and reuse it on all the modules? As I can see, this is boilerplate code right there. What do you think @jumaallan
> 
> Have you seen how the `now-in-android` do it? Using `build-logic` which looks really interesting

Let me have a look into the `now-in-android` setup

jumaallan:
> looks good. My only comment would be to isolate storage & network modules from the core module and core only have the shared items across the modules

I guess we can keep on modularizing as we build - since this is the initial PR anyway

wangerekaharun:
> > looks good. My only comment would be to isolate storage & network modules from the core module and core only have the shared items across the modules
> 
> I guess we can keep on modularizing as we build - since this is the initial PR anyway

Yes, I agree.

jumaallan (approved):

PR #941, opened 2023-02-07
Fix crashes on v1.19.4

jumaallan:


davkutalek (commented):
@jumaallan are you sure about the changes to the account table? I tested that pretty thoroughly and I don't see any crashes related to it, only to the user table.

jumaallan:
The changes actually fix the migration test

Can you try run that locally and see?

davkutalek (approved):

PR #952, opened 2023-02-15
[080223] Migrating channel database to :core:storage module

anselmoalexandre:
Thank you for opening a Pull Request!

- [x] Make sure to open a GitHub issue as a bug/feature request before writing your code! That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests are added/updated (if necessary)
- [x] Ensure the linter passes (`./codeAnalysis` to automatically apply formatting/linting)
- [x] Appropriate docs were updated (if necessary)

Fixes #<issue_number_goes_here> 🦕


anselmoalexandre:
Still waiting for any feedback @jumaallan

jumaallan:
Hey @anselmoalexandre 

When I run with a previous app installed, the app crashes. Here's why - since you modified the `Account` data class, you need to do the following
* bump up the database version to version `53` - then build to let room generate the schema
* write a migration guide in `Migrations.kt` file
* add migration tests inside `MigrationsTest.kt` - you can validate migration or write query to see if the insert works well
* You could write a TypeConverter in `Converters.kt` - to save and read enums from room (optional I guess)

app/src/main/java/com/hover/stax/channels/ChannelUtil.kt
```diff
@@ -0,0 +1,58 @@
...
+                institutionType = ChannelTypes.valueOf(jsonObject.getString("institution_type")),
+                isFavorite = jsonObject.getBoolean("is_favorite")
+            )
+        } catch (e: JSONException) {
+            Timber.d(e.localizedMessage)
+            null
+        }
+    }
+
+    suspend fun load(jsonArray: JSONArray, channelRepository: ChannelRepository, context: Context) {
```
davkutalek, line 37: This should be in `ChannelRepositoryImpl` instead

app/src/main/java/com/hover/stax/channels/ChannelUtil.kt
```diff
@@ -0,0 +1,58 @@
...
+import com.hover.stax.storage.channel.repository.ChannelRepository
+import org.json.JSONArray
+import org.json.JSONException
+import org.json.JSONObject
+import timber.log.Timber
+
+object ChannelUtil {
+    private fun update(jsonObject: JSONObject, context: Context): Channel? {
+        return try {
+            Channel(
```
davkutalek, line 16: This should be a Channel constructor that takes a JSONObject

davkutalek (commented):
Looks good, just the one unnecessary class.

In addition to the migration tests, make sure you do manual test of the migration too. You can check out the previous release (they are tagged eg `v1.19.0`), install it, and then checkout your branch and make sure there are no crashes or errors when you run it.

PR #953, opened 2023-02-20
Bugfix 1.19.6

jumaallan:
Thank you for opening a Pull Request!

- [x] Make sure to open a GitHub issue as a bug/feature request before writing your code! That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests are added/updated (if necessary)
- [x] Ensure the linter passes (`./codeAnalysis` to automatically apply formatting/linting)
- [x] Appropriate docs were updated (if necessary)

This PR fixes the following (read the notes section too) - https://console.firebase.google.com/project/stax-bf03d/crashlytics/app/android:com.hover.stax/issues?time=last-seven-days&versions=1.19.6%20(216)&state=closed&tag=all&sort=eventCount

Fixes #<issue_number_goes_here> 🦕


app/src/main/java/com/hover/stax/transactions/StaxTransaction.kt
```diff
@@ -100,7 +100,7 @@ data class StaxTransaction(
     var accountName: String? = null
 
     constructor(data: Intent, action: HoverAction, contact: StaxContact?, context: Context) : this(
-        data.getStringExtra(TransactionContract.COLUMN_UUID)!!,
+        data.getStringExtra(TransactionContract.COLUMN_UUID) ?: "",
```
davkutalek, line 103: This might cause issues since the UUID is a unique index. I don't think this is where the issue is, because the crash comes when loading from the Hover db, not saving to the Stax db. I can't actually tell what is null from the exception, but I think the Hover db is returning a null transaction for the given uuid rather than the uuid being null.

davkutalek (changes requested):
See comment, I think we may need a null check in  `TransactionDetailsViewModel.loadMessages` instead

app/src/main/java/com/hover/stax/transactions/StaxTransaction.kt
jumaallan: So the fix should be - make the `transaction` nullable inside `hover db` - let the query return a null value?

This will prevent the crash from happening @davkutalek

jumaallan:
As requested by @jimiegha  - add an option to sign in in the settings page

[Screen_recording_20230223_015820.webm](https://user-images.githubusercontent.com/25085146/220782513-56274fe8-43f9-4859-8755-ce451c1edbef.webm)

app/src/main/java/com/hover/stax/transactions/StaxTransaction.kt
davkutalek: yeah, the issue is that the TransactionDao in the SDK is in java, so it is automatically nullable. So yeah, I guess the `TransactionRepo.loadFromHover` call needs to return a nullable Transaction

jumaallan: Do we want to fix that in this fix, or next release? We will have to do a release on the library as well

davkutalek: There is nothing to change in the library since it is java, TransactionRepo is part of Stax

jumaallan: I missed the java part, my bad - makes sense now

jumaallan: Fixed @davkutalek 

Ready for review

davkutalek (approved):

Issue #961, opened 2023-02-22
[Bug]: Pending Status persists for all successful transactions


brysonwaisi:
### Description

Pending status persists for all successful transactions, M-pesa and Safaricom airtime

### Steps to reproduce

1. Open the Stax app.
2. On the Home Screen, tap the refresh icon next to M-Pesa.
3. Enter the MPESA pin.
4. Pay attention to the prompt coming from USSD.

### Expected result

M-Pesa message is returned back, and the pending confirmation status updates to success.

### Actual result

Pending status persists

### Link to screenshots/screen recordings

https://drive.google.com/file/d/17m4ItfLoIARnzBkw5WUObNA6mJx9DXaU/view?usp=share_link

### Device name

Samsung Galaxy A20s

### Android OS

11

### App version

1.19.6

### Config version

20230216140008

### Is your phone dual SIM or single SIM?

Dual SIM

### Which telco are you using for this transaction?

Safaricom

### Stax device ID or email address

brysonnyamwange@gmail.com

PR #989, opened 2023-03-26
Fix Pending Status persists for all successful transactions

GibsonRuitiari:
This PR fixes the pending status that persists after all successful transactions.

Fixes #[965](https://github.com/UseHover/Stax/issues/965) and  #[961](https://github.com/UseHover/Stax/issues/961)🦕

### DIFF
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="CompilerConfiguration">
-    <bytecodeTargetLevel target="11" />
+    <bytecodeTargetLevel target="17" />
   </component>
 </project>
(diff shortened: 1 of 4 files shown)


jumaallan (approved):
Interesting fix 👏🏽

GibsonRuitiari:
A little explanation for future persons :-)

When your app components create different instances of a Dao, the instances are not synchronized. Hence you might find that one instance of your Dao is busy updating the database whereas another is busy reading the database. So the one reading the database might in most cases get stale data.

One solution could be providing a singleton instance of your Dao and injecting the instance to all your components. Since you only have one single instance, from how room is designed, it will be up-to-date with the current data in the table.

Another solution could be forcing room to invalidate table instances and subsequently invalidate your Dao instances upon an update so that any component using your Dao will get an updated instance of it. Behind the scenes what room does is it uses a service that tracks the database changes and updates the instances of your Dao's. So anytime your data changes, the instances are updated and the Dao's are always synchronized.

I have gone with the second solution since it is the easier one.

jumaallan:
> A little explanation for future persons :-)
> 
> When your app components create different instances of a Dao, the instances are not synchronized. Hence you might find that one instance of your Dao is busy updating the database whereas another is busy reading the database. So the one reading the database might in most cases get stale data.
> 
> One solution could be providing a singleton instance of your Dao and injecting the instance to all your components. Since you only have one single instance, from how room is designed, it will be up-to-date with the current data in the table.
> 
> Another solution could be forcing room to invalidate table instances and subsequently invalidate your Dao instances upon an update so that any component using your Dao will get an updated instance of it. Behind the scenes what room does is it uses a service that tracks the database changes and updates the instances of your Dao's. So anytime your data changes, the instances are updated and the Dao's are always synchronized.
> 
> I have gone with the second solution since it is the easier one.

Thanks for this 👏🏽

as7r44:
Hello guys,
was this fix merged with master branch or tested and working fine with you ?
As still I am facing this challenge of pending status on successfull transactions

PR #1002, opened 2023-04-17
Cleanup android editable to android inputType to use valid phone number

### LINKED ISSUE #1001: [Bug]: A User can input invalid inputs on the Contact Phone Number TextField
### Description

When a user inputs the recipient phone number while sending money, the user can input invalid characters e.g. Alphabet characters and even write unlimited number characters on the text field. This should not be the case. The user should get a warning to input correct phone number before proceeding to the next steps.

### Steps to reproduce

1. Open Stax app. 
2. Tap Send Money
3. On the Contact Phone Number text field, write any incorrect number or any text input instead of a valid number 

### Expected result

User should get a warning before they proceed to the next steps. They should get a warning like "Invalid phone number input a valid number"

### Actual result

No warning. The user can proceed but they will eventually get a failed message after they tap continue and tap send money now.

### Link to screenshots/screen recordings

Screenshot_20230411_174147_Stax (1)

### Suggested solution

_No response_

### Device name

Samsung S22+

### Android OS

13

### App version

1.19.8

### Config version

_No response_

### Is your phone dual SIM or single SIM?

Dual SIM

### Which telco are you using for this transaction?

Safaricom

### Stax device ID or email address

langat.caleb95@gmail.com

Mzazi25:
This PR updates android:editable which has been deprecated to use  android: inputType on the Contact phone number TextField. This would ensure the user would enter a valid phone number. 

- [x ] Make sure to open a GitHub issue as a bug/feature request before writing your code! That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests are added/updated (if necessary)
- [ ] Ensure the linter passes (`./codeAnalysis` to automatically apply formatting/linting)
- [ ] Appropriate docs were updated (if necessary)

Fixes #1001  🦕

### DIFF
--- a/app/src/main/java/com/hover/stax/views/StaxDropdownLayout.kt
+++ b/app/src/main/java/com/hover/stax/views/StaxDropdownLayout.kt
@@ -34,7 +34,7 @@ open class StaxDropdownLayout(context: Context, attrs: AttributeSet) : AbstractS
     lateinit var autoCompleteTextView: AutoCompleteTextView
     private var hint: String? = null
     private var defaultText: String? = null
-    private var editable = false
+    private var inputType:String? = null
     private var imeOptions = 0
 
     @CallSuper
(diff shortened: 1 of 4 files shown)


davkutalek:
@Mzazi25 the `StaxDropdownLayout` is used by more than just the contact input, so this likely breaks some of its usages. The use of editable is to change it between a freeform input with autocomplete options and a regular dropdown with only preset options. Even the contact case it is more complicated: if a contact is chosen from the address book then it shows the contact's name in the input field.

@jumaallan is in the middle of converting the entire flow to compose anyway, so probably not worth the time to try to sort out the different use cases or get rid of the depreciated `editable` option

PR #1013, opened 2023-05-04
Migrated db entities and repo 

jumaallan:
This also replaces koin setup with hilt so the PR might get really huge


jumaallan:
This PR breaks everything - we need to fix DI on data module next

